PR #22, opened 2022-07-11
Use default stiffness for spring()

zsoltk:
## Description

Remove overridden spring specs due to compose interrupt transition behaviour.
...

## Check list

- [x] `CHANGELOG.md` not updated
- [x] Documentation not updated


KovalevAndrey:
> Remove overridden spring specs due to compose interrupt transition behaviour.

Not sure I understand it. They only check if the provided spec is [instance of Spring class](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/animation/animation-core/src/commonMain/kotlin/androidx/compose/animation/core/Transition.kt;l=551)

KovalevAndrey:
fixed here 
https://github.com/bumble-tech/appyx/pull/24

zsoltk:
My bad, the issue was with #24.

PR #30, opened 2022-07-12
Merge snapshot.yml into build.yml and use "cache-read-only: true"

CherryPerry:
## Description

`cache-read-only: true` to not override cache from main build.

## Check list


.github/workflows/build.yml
```diff
@@ -39,3 +40,28 @@ jobs:
           python-version: '3.x'
       - run: pip install mkdocs-material
       - run: mkdocs build --strict
+
+  publish-snapshot:
+    name: Publish snapshot
+    runs-on: ubuntu-latest
+    needs: [ build ]
+    if: github.ref == 'refs/heads/main'
```
CherryPerry, line 48: Only on `main` branch.

LachlanMcKee (approved):

PR #32, opened 2022-07-14
BackPressHandler plugin

CherryPerry:
## Description

[Currently](https://github.com/bumble-tech/appyx/blob/main/core/src/main/java/com/bumble/appyx/core/plugin/Plugins.kt#L31) `BackPressHandler` is not an actual plugin and is manually supported only for a routing source. In this PR I tried to implement `BackPressHandler` that is available for any component in a node.

Also in this PR i tried to support [predictive back gesture](https://developer.android.com/about/versions/13/features/predictive-back-gesture) feature of Android 13. As you can see on the attached video it works as expected.

### AndroidX

In this PR I implement `BackPressHandler` as a provider of AndroidX `OnBackPressedCallback`. It allows more seamless integration with other AndroidX components but has the following issues:
- We are one step further away from multi platform support and it is Android only library.
- Client has to work with not the best of class API, but still more familiar than some custom API.
- `OnBackPressedCallback.isEnabled` is not reactive/observable, so we need to manually synchronise internal state of an owner component with this property like `feature.state.subscribe { callback.isEnabled = state.someCheck }`.
 - In case of custom API it is possible to create more interesting API like
```kotlin
interface OnBackPressedCallback {
   // flow here instead of simple boolean property
   // to be able to easily map internal state into boolean state in a reactive way 
   val isEnabled: StateFlow<Boolean>
   fun dispatchBackPressed()
}
```
 - But in this case we are forcing clients (including us) to use `StateFlow` in their internal components (like interactors) instead of RxJava for example. But we are already doing it in `RoutingSource`.
- We can't keep current API of `fun onBackPressed()` because we need to declare if we support going back or not first, so AndroidX and Android integration works fine. This change also allowed me to remove one property from `RoutingSource`.
- By using AndroidX API directly we are guarding ourselves from potential changes like predictive back button in the future. 

### Registration order

In the current implementation and in this pull request we are relying on Compose to register `OnBackPressedCallback`. Invocation order of `OnBackPressedCallback` of different nodes in the same parent is defined by the order of Composable functions. If a node is out of a composition, it won't handle back presses. Implementation details is [here](https://github.com/bumble-tech/appyx/blob/main/core/src/main/java/com/bumble/appyx/core/node/ParentNode.kt#L217).

Most likely this problem is out of scope of this pull request 

## Check list

[predictive-support.webm](https://user-images.githubusercontent.com/9081555/179012451-4b4cbfd1-4a67-4b0d-b6db-a7cfb8548f6e.webm)

### DIFF
--- a/core/src/androidTest/java/com/bumble/appyx/core/plugin/BackPressHandlerTest.kt
+++ b/core/src/androidTest/java/com/bumble/appyx/core/plugin/BackPressHandlerTest.kt
@@ -1,0 +1,145 @@
+package com.bumble.appyx.core.plugin
+
+import android.app.Activity
+import androidx.activity.OnBackPressedCallback
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.test.espresso.Espresso
+import androidx.test.platform.app.InstrumentationRegistry
+import com.bumble.appyx.core.composable.Children
+import com.bumble.appyx.core.modality.BuildContext
+import com.bumble.appyx.core.node.ParentNode
+import com.bumble.appyx.core.node.node
+import com.bumble.appyx.debug.Appyx
+import com.bumble.appyx.routingsource.backstack.BackStack
+import com.bumble.appyx.routingsource.backstack.activeRouting
+import com.bumble.appyx.routingsource.backstack.operation.push
+import com.bumble.appyx.testing.ui.rules.AppyxTestRule
+import org.hamcrest.MatcherAssert.assertThat
+import org.hamcrest.Matchers.equalTo
+import org.hamcrest.Matchers.instanceOf
+import org.junit.After
+import org.junit.Rule
+import org.junit.Test
+
+class BackPressHandlerTest {
+
+    private var onBackPressedHandled = false
+    private var backPressHandler: BackPressHandler = object : BackPressHandler {
+        override val onBackPressedCallback: OnBackPressedCallback =
+            object : OnBackPressedCallback(true) {
+                override fun handleOnBackPressed() {
+                    onBackPressedHandled = true
+                }
+            }
+    }
+
+    @get:Rule
+    val rule = AppyxTestRule(launchActivity = false) { buildContext ->
+        TestParentNode(buildContext = buildContext, plugin = backPressHandler)
+    }
+
+    @After
+    fun after() {
+        Appyx.exceptionHandler = null
+    }
+
+    @Test
+    fun routing_handles_back_press_when_plugin_has_disabled_listener() {
+        rule.start()
+        runOnMainSync {
+            backPressHandler.onBackPressedCallback!!.isEnabled = false
+            rule.node.backStack.push(TestParentNode.Routing.ChildB)
+        }
+
+        Espresso.pressBack()
+        Espresso.onIdle()
+
+        assertThat(rule.node.backStack.activeRouting, equalTo(TestParentNode.Routing.ChildA))
+        assertThat(onBackPressedHandled, equalTo(false))
+    }
+
+    @Test
+    fun custom_plugin_handles_back_press_before_routing() {
+        rule.start()
+        runOnMainSync { rule.node.backStack.push(TestParentNode.Routing.ChildB) }
+
+        Espresso.pressBack()
+        Espresso.onIdle()
+
+        assertThat(rule.node.backStack.activeRouting, equalTo(TestParentNode.Routing.ChildB))
+        assertThat(onBackPressedHandled, equalTo(true))
+    }
+
+    @Test
+    fun activity_is_closed_when_nobody_can_handle_back_press() {
+        rule.start()
+        runOnMainSync {
+            backPressHandler.onBackPressedCallback!!.isEnabled = false
+        }
+
+        Espresso.pressBackUnconditionally()
+        Espresso.onIdle()
+
+        assertThat(rule.activityResult.resultCode, equalTo(Activity.RESULT_CANCELED))
+    }
+
+    @Test
+    fun reports_incorrect_handler() {
+        var exception: Exception? = null
+        Appyx.exceptionHandler = { exception = it }
+        backPressHandler = object : BackPressHandler {
+            override val onBackPressedCallback: OnBackPressedCallback =
+                object : OnBackPressedCallback(true) {
+                    override fun handleOnBackPressed() {
+                    }
+                }
+
+            override val onBackPressedCallbackList: List<OnBackPressedCallback> = listOf(
+                onBackPressedCallback, object : OnBackPressedCallback(true) {
+                    override fun handleOnBackPressed() {
+                    }
+                }
+            )
+        }
+        rule.start()
+
+        Espresso.onIdle()
+
+        assertThat(exception, instanceOf(IllegalStateException::class.java))
+    }
+
+    private fun runOnMainSync(runner: Runnable) {
+        InstrumentationRegistry.getInstrumentation().runOnMainSync(runner)
+    }
+
+    class TestParentNode(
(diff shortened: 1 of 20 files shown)


core/src/main/java/com/bumble/appyx/core/node/Node.kt
```diff
@@ -95,16 +99,37 @@ abstract class Node(
...
+                backPressHandlerPlugins.forEach { plugin ->
+                    plugin.onBackPressedCallbacks.forEach { it.remove() }
+                }
+            }
+        }
     }
 
+    // BackPressHandler is correct when only one of its properties is implemented.
+    private fun BackPressHandler.isCorrect(): Boolean =
+        onBackPressedCallback != null && listOf(onBackPressedCallback) == onBackPressedCallbacks ||
```
LachlanMcKee, line 130: `onBackPressedCallback == onBackPressedCallbacks.firstOrNull()`?

CherryPerry, line 130: Split into 2 variable.

core/src/main/java/com/bumble/appyx/core/plugin/Plugins.kt
```diff
@@ -21,16 +22,35 @@ interface NodeLifecycleAware : Plugin {
...
+ * In case if both implemented, [onBackPressedCallback] will be ignored.
+ * There is runtime check in [Node] to verify correctness.
+ */
 interface BackPressHandler : Plugin {
-    fun onBackPressed()
+
+    val onBackPressedCallback: OnBackPressedCallback? get() = null
+
+    /** It is impossible to combine multiple [OnBackPressedCallback] into the single one, they are not observable. */
+    val onBackPressedCallbacks: List<OnBackPressedCallback>
```
CherryPerry, line 44: `onBackPressedCallbacksList`?

core/src/main/java/com/bumble/appyx/core/node/Node.kt
CherryPerry: If `onBackPressedCallbacks` has more than one element, then it means that you override both.

LachlanMcKee: Was suggesting avoiding creating an object just for the comparison

CherryPerry: `onBackPressedCallbacks.size == 1 && onBackPressedCallbacks[0] == onBackPressedCallback`

core/src/main/java/com/bumble/appyx/core/plugin/Plugins.kt
```diff
@@ -21,16 +22,35 @@ interface NodeLifecycleAware : Plugin {
...
+ * Implementing class can handle back presses via [OnBackPressedCallback].
+ *
+ * Implement either [onBackPressedCallback] or [onBackPressedCallbacks], not both.
+ * In case if both implemented, [onBackPressedCallback] will be ignored.
+ * There is runtime check in [Node] to verify correctness.
+ */
 interface BackPressHandler : Plugin {
-    fun onBackPressed()
+
+    val onBackPressedCallback: OnBackPressedCallback? get() = null
```
SamuPS, line 41: What do you think about just removing this property? 
If we do so we can just remove the checks in the node and there would be no risk isn't it?
Probably the common user will not usually work around this interface as we are already hiding this complexity from  the BaseRoutingSource, so i will say that we can avoid  so i think that we can be less kind with this api and force extensions to pass a list if the want to override.

What do you think?

CherryPerry: It might be used not only for routing sources but also for interactors.

core/src/main/java/com/bumble/appyx/core/node/Node.kt
```diff
@@ -95,14 +99,39 @@ abstract class Node(
...
+        }
+    }
+
+    // BackPressHandler is correct when only one of its properties is implemented.
+    private fun BackPressHandler.isCorrect(): Boolean {
+        val listIsOverriddenOrPluginIgnored = onBackPressedCallback == null
+        val onlySingleCallbackIsOverridden = onBackPressedCallback != null &&
+                onBackPressedCallbackList.size == 1 &&
+                onBackPressedCallbackList[0] == onBackPressedCallback
+        return onlySingleCallbackIsOverridden || listIsOverriddenOrPluginIgnored
```
LachlanMcKee, line 134: should this be `return onlySingleCallbackIsOverridden != listIsOverriddenOrPluginIgnored`?

Otherwise if they are both true it will still be deemed correct

core/src/androidTest/java/com/bumble/appyx/core/plugin/BackPressHandlerTest.kt
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    @get:Rule
+    val rule = AppyxTestRule(launchActivity = false) { buildContext ->
+        TestParentNode(buildContext = buildContext, plugin = backPressHandler)
+    }
+
+    @After
+    fun after() {
+        Appyx.exceptionHandler = null
```
LachlanMcKee, line 44: Should this also be set in a before block? In case the exception handler was altered in a different test class?

core/src/main/java/com/bumble/appyx/core/node/Node.kt
```diff
@@ -104,14 +108,40 @@ abstract class Node(
...
+            onDispose {
+                backPressHandlerPlugins.forEach { plugin ->
+                    plugin.onBackPressedCallbackList.forEach { it.remove() }
+                }
+            }
+        }
+    }
+
+    // BackPressHandler is correct when only one of its properties is implemented.
+    private fun BackPressHandler.isCorrect(): Boolean {
```
LachlanMcKee, line 139: companion object?

core/src/main/java/com/bumble/appyx/core/node/Node.kt
```diff
@@ -104,14 +108,40 @@ abstract class Node(
...
 
-    /** Derived classes can declare functional (non-ui) Composable blocks before [View()] is invoked. */
     @Composable
-    protected open fun DerivedSetup() {
+    private fun HandleBackPress() {
+        // reversed order because we want direct order, but onBackPressedDispatcher invokes them in reversed order
+        val backPressHandlerPlugins = remember { plugins.filterIsInstance<BackPressHandler>().reversed() }
+        val dispatcher = LocalOnBackPressedDispatcherOwner.current?.onBackPressedDispatcher
+        DisposableEffect(dispatcher) {
+            backPressHandlerPlugins.forEach { plugin ->
```
LachlanMcKee, line 122: If dispatcher is null, do we need to iterate over the list?

LachlanMcKee (commented):
Could you add an example in the Sandbox? I can't figure out how it should be used in a real app/feature. The Android Test does not make it clear how straightforward it is.

core/src/androidTest/java/com/bumble/appyx/core/plugin/BackPressHandlerTest.kt
CherryPerry: I think it is others test responsibility to clean the static state. Cleaning the static state everywhere seems hard to implement and maintain.

LachlanMcKee (approved):

core/src/androidTest/java/com/bumble/appyx/core/plugin/BackPressHandlerTest.kt
LachlanMcKee: fair enough

LachlanMcKee:
Thanks for updating the example!

PR #35, opened 2022-07-17
Issue #34 Add unit and instrumentation tests to github action

### LINKED ISSUE #34: Add AndroidTest execution to Github actions
Currently it is possible for someone to merge to main without running the AndroidTests within the Sandbox module (and potentially future modules). This could mean that the tests will break silently.

LachlanMcKee:
## Description

Ensure that unit tests and instrumentation tests are run as part of the PR

Fixes #34 

## Check list

- [x] I do not need to update `CHANGELOG.md`.
- [x] I do not need to update the documentation.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -54,8 +54,9 @@ jobs:
           java-version: '11'
       - uses: gradle/wrapper-validation-action@v1
       - name: Unit tests
-        run: ./gradlew testDebugUnitTest
-        working-directory: .
+        uses: gradle/gradle-build-action@v2
+        with:
+          arguments: testDebugUnitTest
       - name: Upload failed unit test artifacts
         if: failure()
         uses: actions/upload-artifact@v3
@@ -96,6 +97,8 @@ jobs:
           emulator-options: -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
           disable-animations: false
           script: echo "Generated AVD snapshot for caching."
+      - name: Setup Gradle
+        uses: gradle/gradle-build-action@v2
       - name: Instrumentation tests
         uses: reactivecircus/android-emulator-runner@v2
         with:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -42,6 +42,80 @@ jobs:
             -Psonatype.username=${{ secrets.SONATYPE_USERNAME }}
             -Psonatype.password=${{ secrets.SONATYPE_PASSWORD }}
 
+  unit-tests:
+    name: Unit tests
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-java@v3
+        with:
+          distribution: 'zulu'
+          java-version: '11'
+      - uses: gradle/wrapper-validation-action@v1
+      - name: Unit tests
+        run: ./gradlew testDebugUnitTest
+        working-directory: .
+      - name: Upload failed unit test artifacts
+        if: failure()
+        uses: actions/upload-artifact@v3
+        with:
+          name: unit-test-failures
+          path: '**/build/reports'
+
+  instrumentation-tests:
+    name: Instrumentation tests
+    runs-on: macOS-latest
+    timeout-minutes: 20
+    strategy:
+      fail-fast: false
+      matrix:
+        api-level: [31]
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-java@v3
+        with:
+          distribution: 'zulu'
+          java-version: '11'
+      - uses: gradle/wrapper-validation-action@v1
+      - name: AVD cache
+        uses: actions/cache@v3
+        id: avd-cache
+        with:
+          path: |
+            ~/.android/avd/*
+            ~/.android/adb*
+          key: avd-${{ matrix.api-level }}
+      - name: create AVD and generate snapshot for caching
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -42,6 +42,78 @@ jobs:
...
+        run: ./gradlew testDebugUnitTest
+        working-directory: .
+      - name: Upload failed unit test artifacts
+        if: failure()
+        uses: actions/upload-artifact@v3
+        with:
+          name: unit-test-failures
+          path: '**/build/reports'
+
+  instrumentation-tests:
```
LachlanMcKee, line 66: We might want to make this depend on the `build` job running successfully, as it can take quite a lot longer than the other jobs (and I believe you can only use on MacOS agent at a time - i.e. multiple PRs would have to wait for the agent to be free). 

However that would make the PR take slightly longer to be ready

.github/workflows/build.yml
```diff
@@ -42,6 +42,80 @@ jobs:
...
+    name: Instrumentation tests
+    runs-on: macOS-latest
+    timeout-minutes: 20
+    strategy:
+      fail-fast: false
+      matrix:
+        api-level: [31]
```
LachlanMcKee, line 73: No AVD exists for API 32 without Google APIs

https://gist.github.com/alvr/8db356880447d2c4bbe948ea92d22c23

If we want to use API-level 32 (which is the compile target), we would need to set `target: google_apis`. That would then mean the library could hypothetically not work on stock Android.

.github/workflows/build.yml
```diff
@@ -42,6 +42,83 @@ jobs:
...
+        uses: reactivecircus/android-emulator-runner@v2
+        with:
+          api-level: ${{ matrix.api-level }}
+          arch: x86_64
+          force-avd-creation: false
+          emulator-options: -no-snapshot-save -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
+          disable-animations: true
+          script: |
+            adb logcat > logcat.out &
+            ./gradlew :sandbox:connectedCheck --stacktrace
```
LachlanMcKee, line 112: It seems to be significantly faster to specify the module rather than execute `connectedCheck`. Annoyingly without specifying the module it tries to build the androidTest tasks for all the modules even though they don't have any tests.

I was trying to find a way to potentially remove the connectedCheck task from those modules to avoid them being run, but I'm struggling to find the right API.

Any thoughts? It potentially adds 4-5 minutes of extra waiting.

.github/workflows/build.yml
CherryPerry: `build` depends on `test*UnitTest` and `lint` tasks, you can launch locally to verify. No need to have a separate config for it.
Please just add `actions/upload-artifact@v3` to the main config.

.github/workflows/build.yml
CherryPerry: 1. AFAIK 32 is a sub-release with low level of adoption because it is targeted only for foldable devices and tablets.
2. Testing on 31 is fine too.
3. 33 release is [soon](https://android-developers.googleblog.com/2022/07/Final-Android-13-Beta-update-official-release-is-next.html).

So just use 31 (because non-Google images are less bloated.

.github/workflows/build.yml
CherryPerry: Merge with `build` job so you have much less to compile for `connectedCheck`.

> `--stacktrace`

Not sure why do you need it.

.github/workflows/build.yml
```diff
@@ -42,6 +42,83 @@ jobs:
...
+          name: unit-test-failures
+          path: '**/build/reports'
+
+  instrumentation-tests:
+    name: Instrumentation tests
+    runs-on: macOS-latest
+    timeout-minutes: 20
+    strategy:
+      fail-fast: false
+      matrix:
```
CherryPerry, line 73: Do you want a matrix here? Only one API level is used.
AFAIK we do not use any Android OS dependent APIs. If we even use them, we do not have tests to cover a behaviour or just relying to AndroidX compatibility libraries.

CherryPerry (changes requested):
Just throw everything into `build` job. Otherwise you need to spend time to setup Gradle, Gradle cache, dependencies and to compile everything each time on every VM.

.github/workflows/build.yml
LachlanMcKee: Sounds good! thanks :)

.github/workflows/build.yml
LachlanMcKee: good catch. I was using this before due to a problem. will remove it

.github/workflows/build.yml
LachlanMcKee: happy to remove it. thought it might be wise just in case we decided to go down that path.

LachlanMcKee:
@CherryPerry I think it would be wise to still have the instrumentation tests in a separate job. The recommendation is to use a macOS machine (due to HAXM) for Android tests.

You only get one macOS agent per project (unless you pay for it), so it would prevent us from running multiple PRs in parallel

CherryPerry:
There is no easy way to reuse a cache between different jobs (the main job on Linux and UI tests job on macOS), so let's keep a separate job for android tests. Use `connectedTest` without module specification. The issue with compilation of unnecessary modules can be solved with custom Gradle config, I can implement it in the future if the current state will become a problem.

LachlanMcKee:
@CherryPerry done :)

CherryPerry (approved):

Issue #39, opened 2022-07-22
Implement workflows in order to support navigation via deeplinks


KovalevAndrey:
At the moment navigation via deeplinks in not supported. Similarly to RIBs we need to implement workflows in order to support it

PR #40, opened 2022-07-22
Implement workflows + deepnlink navigation example

KovalevAndrey:
## Description

Fixes  https://github.com/bumble-tech/appyx/issues/39

Implemented workflows for deeplinks navigation + examples of how to use it.

## Check list


sandbox/src/main/java/com/bumble/appyx/sandbox/client/workflow/WorkflowExampleActivity.kt
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    override fun onDestroy() {
+        super.onDestroy()
+        job?.cancel()
+    }
+
+    private val workflowFactory: (Intent) -> Job? = {
+        when {
+            // adb shell am start -a "android.intent.action.VIEW" -d "appyx-sample://workflow1"
```
LachlanMcKee, line 37: These would be a good candidate for androidTests

KovalevAndrey: Yes, tests will be added once we agree on the api

core/src/main/java/com/bumble/appyx/core/node/Node.kt
```diff
@@ -80,6 +84,13 @@ abstract class Node(
         });
     }
 
+    protected suspend inline fun <reified T> executeWorkflow(
+        crossinline action: () -> Unit
+    ): T = withContext(lifecycleScope.coroutineContext + Dispatchers.Main) {
```
CherryPerry, line 89: `Lifecycle.lifecycleScope` is already `Main`.

core/src/main/java/com/bumble/appyx/core/node/ParentNode.kt
```diff
@@ -212,6 +221,48 @@ abstract class ParentNode<Routing : Any>(
         }
     }
 
+    protected suspend inline fun <reified T> attachWorkflow(
```
CherryPerry, line 224: Please add tests.

core/src/main/java/com/bumble/appyx/core/node/ParentNode.kt
```diff
@@ -212,6 +221,48 @@ abstract class ParentNode<Routing : Any>(
...
+                    "Last child is: [${childNodes.lastOrNull()}]. " +
+                    "All children: $childNodes"
+            )
+        } else {
+            childNodesOfExpectedType.last()
+        }
+    }
+
+    protected suspend inline fun <reified T : Node> waitForChildAttached(): T =
+        withContext(lifecycleScope.coroutineContext + Dispatchers.Main) {
```
CherryPerry, line 244: `children` is a `Flow` so I think you can just subscribe to it and take the first result.

sandbox/src/main/java/com/bumble/appyx/sandbox/client/workflow/WorkflowExampleActivity.kt
```diff
@@ -0,0 +1,92 @@
...
+        super.onCreate(savedInstanceState)
+        setContent {
+            AppyxSandboxTheme {
+                // A surface container using the 'background' color from the theme
+                Surface(color = MaterialTheme.colors.background) {
+                    Column {
+                        NodeHost(integrationPoint = integrationPoint) {
+                            RootNode(
+                                buildContext = it,
+                                plugins = listOf(object : ParenNodeBuilt<RootNode> {
```
CherryPerry, line 61: `NodeAware<RootNode>` does not help?

core/src/main/java/com/bumble/appyx/core/node/ParentNode.kt
```diff
@@ -212,6 +221,48 @@ abstract class ParentNode<Routing : Any>(
...
+                    "Child count: ${childNodes.size}. " +
+                    "Last child is: [${childNodes.lastOrNull()}]. " +
+                    "All children: $childNodes"
+            )
+        } else {
+            childNodesOfExpectedType.last()
+        }
+    }
+
+    protected suspend inline fun <reified T : Node> waitForChildAttached(): T =
```
LachlanMcKee, line 243: Should we have an optional timeout in case the child is never attached?

core/src/main/java/com/bumble/appyx/core/node/Node.kt
```diff
@@ -80,6 +84,13 @@ abstract class Node(
         });
     }
 
+    protected suspend inline fun <reified T> executeWorkflow(
```
LachlanMcKee, line 87: How does this method relate to `attachWorkflow`? Workflow seems to be used in attachWorkflow as another way of adding multiple children (like a deeplink)

I'm struggling to see how they relate other than at what point in the code this can execute (as you said `Workflows can be executed only after root node onBuilt has been executed - when childNodeLifecycleManager.launch() is called`)

core/src/main/java/com/bumble/appyx/core/node/Node.kt
KovalevAndrey: it uses `Main.immediate` which can change executing order. `attachWorkflow` doesn't work without `Dispatchers.Main` because right after after calling action we search for a certain element. Because the order is changed children won't be synced at this time. 

Execute workflow works without explicitly setting `Dispatchers.Main` but for the sake of consistency I'd keep it

core/src/main/java/com/bumble/appyx/core/node/ParentNode.kt
KovalevAndrey: I think we shouldn't. It's fine for the job to hang around as long as needed.

core/src/main/java/com/bumble/appyx/core/node/Node.kt
KovalevAndrey: these 2 methods are independent:
1. `attachWorkflow` can be only called on a parent node as it attaches a child node
2. `executeWorkflow` just performs an action.

We only need to wait for `RootNode` to be built in order to execute a workflow. It doesn't work as is because at the moment there's no callback from composition to know when the root node has been actually composed and built

KovalevAndrey: Naming is matching [RIBs](https://github.com/badoo/RIBs/blob/master/libraries/rib-rx2/src/main/java/com/badoo/ribs/rx2/workflows/RxWorkflowNode.kt)

sandbox/src/main/java/com/bumble/appyx/sandbox/client/workflow/WorkflowExampleActivity.kt
KovalevAndrey: actually it does with Main dispatchers

core/src/main/java/com/bumble/appyx/core/node/Node.kt
KovalevAndrey: Removed. Now we perform an action and waiting for the children to be in sync with routing source. After 1 second throw an exception if desired Node was not found

sandbox/src/main/java/com/bumble/appyx/sandbox/client/workflow/WorkflowExampleActivity.kt
KovalevAndrey: removed

core/src/main/java/com/bumble/appyx/core/node/ParentNode.kt
zsoltk: I like the idea of an optional timeout, as executing the original workflow after say half an hour could be unwanted and outdated. It's fine as a separate PR too.

KovalevAndrey: > I like the idea of an optional timeout, as executing the original workflow after say half an hour could be unwanted and outdated. It's fine as a separate PR too.

Why don't let clients have full control over workflows? If they need time-out they can implement it with a one-liner?

sandbox/src/main/java/com/bumble/appyx/sandbox/client/workflow/GrandChildNodeB.kt
```diff
@@ -0,0 +1,115 @@
...
+import com.bumble.appyx.core.node.Node
+import com.bumble.appyx.core.node.ParentNode
+import com.bumble.appyx.routingsource.backstack.BackStack
+import com.bumble.appyx.routingsource.backstack.operation.push
+import com.bumble.appyx.sandbox.client.workflow.ChildNodeB.Routing
+import com.bumble.appyx.sandbox.client.workflow.ChildNodeB.Routing.ChildA
+import com.bumble.appyx.sandbox.client.workflow.ChildNodeB.Routing.ChildB
+import kotlinx.parcelize.Parcelize
+
+class ChildNodeB(
```
zsoltk, line 28: Filename doesn't match class name

sandbox/src/main/java/com/bumble/appyx/sandbox/client/workflow/WorkflowExampleActivity.kt
```diff
@@ -0,0 +1,95 @@
...
+                                buildContext = it,
+                                plugins = listOf(object : NodeAware<RootNode> {
+                                    override fun init(node: RootNode) {
+                                        rootNode = node
+                                        handleDeepLink(intent = intent)
+                                    }
+
+                                    override val node: RootNode
+                                        get() = rootNode
+                                })
```
zsoltk, line 69: I wonder if this would make sense to provide a dedicated library plugin for?

sandbox/src/main/java/com/bumble/appyx/sandbox/client/workflow/WorkflowExampleActivity.kt
```diff
@@ -0,0 +1,95 @@
...
+                }
+            }
+        }
+    }
+
+    private fun executeWorkflow1(): Job {
+        return lifecycleScope.launch {
+            rootNode
+                .waitForChildBAttached()
+                .attachChildBCoroutine()
```
zsoltk, line 82: Why? The wait method already attaches

sandbox/src/main/java/com/bumble/appyx/sandbox/client/workflow/WorkflowExampleActivity.kt
```diff
@@ -0,0 +1,95 @@
...
+        }
+    }
+
+
+    private fun executeWorkflow2(): Job {
+        return lifecycleScope.launch {
+            rootNode
+                .attachChildBCoroutine()
+                .attachChildBCoroutine()
```
zsoltk, line 91: Why twice?

CherryPerry:
This API forces coroutine usage (we already have `Router` that does the same), let's discuss later what to do with it.

core/src/main/java/com/bumble/appyx/core/node/ParentNode.kt
```diff
@@ -284,6 +321,7 @@ abstract class ParentNode<Routing : Any>(
     }
 
     companion object {
+        const val AttachWorkFlowSyncTimeout = 1000L
```
CherryPerry, line 324: ALL_CAPS.
Also 1 seconds looks too fast, maybe 10?

sandbox/src/main/java/com/bumble/appyx/sandbox/client/workflow/WorkflowExampleActivity.kt
KovalevAndrey: This is childB of childB (not the same action)

sandbox/src/main/java/com/bumble/appyx/sandbox/client/workflow/WorkflowExampleActivity.kt
KovalevAndrey: This pushes grandchild of root node, I can rename to avoid confusion

LachlanMcKee:
@CherryPerry perhaps these could be extension functions? That could avoid coupling the API with coroutines potentially

core/src/main/java/com/bumble/appyx/core/node/ParentNode.kt
```diff
@@ -212,6 +218,37 @@ abstract class ParentNode<Routing : Any>(
...
+                "Check that your action actually results in the expected child. "
+        )
+    }
+
+    protected suspend inline fun <reified T : Node> waitForChildAttached(): T =
+        suspendCancellableCoroutine { continuation ->
+            lifecycleScope.launch {
+                children.collect { childMap ->
+                    val childNodes = childMap.entries.mapNotNull { it.value.nodeOrNull }
+                    val childNodesOfExpectedType = childNodes.filterIsInstance<T>()
```
CherryPerry, line 242: It can be implemented without creation of 2 new lists:
```kotlin
childMap.entries.findLast { it is T } as T?
```

Also `last()` does not really makes sense because `Map.entries` is a `Set`.

core/src/test/java/com/bumble/appyx/core/node/NodeTest.kt
```diff
@@ -0,0 +1,56 @@
...
+    @get:Rule
+    val mainDispatcherRule = MainDispatcherRule()
+
+    private val testScope = TestScope(UnconfinedTestDispatcher())
+
+    @Test
+    fun `executeWorkflow WHEN called THEN executes action returns current node`() =
+        testScope.runTest {
+            val node = TestNode()
+            assertThat(node.status, IsNull())
```
CherryPerry, line 32: How did you get `hamcrest` here? It should be available only in UI tests (as Espresso dependency).

core/src/main/java/com/bumble/appyx/core/node/ParentNode.kt
KovalevAndrey: synchronisation of children with routing source happens on main thread, I think 1 second should be more than enough for the child to appear in the Parent's Node and we don't have to wait for additional 9 seconds in case it wasn't attached properly

LachlanMcKee: It's definitely possible for the main thread to be blocked for more than 1 second (ANRs)

core/src/main/java/com/bumble/appyx/core/node/ParentNode.kt
LachlanMcKee: Perhaps we should add some documentation to the method to make it clear that it will potentially block forever and that they should add their own timeout?

A lot of Java APIs that have await style methods usually have a timeout overload that makes it fairly obvious that not passing in a timeout value would cause it to wait forever

core/src/main/java/com/bumble/appyx/core/node/ParentNode.kt
CherryPerry: Waiting is suspended, no main thread blocks. IMHO 1 seconds still looks too fast, the main thread might be busy with something else, and wait function does not take it into account.

KovalevAndrey: Okay, I'll increase the timeout, but 10 is way too long in my opinion, should I make it 5 at least? So we'll fail faster

LachlanMcKee: @CherryPerry I was aware that this method wouldn't block the main thread, I was making the same point that you just made. The main thread could be busy for more than 1 second, causing this to time out.

For an ANR to occur, the UI thread is blocked for more than 5 seconds (more in some cases). There are quite a lot of apps that have more than 0% ANRs, so it is possible that this will time out. In this circumstance what would be the expectation from the developer?

The workflow would be cancelled? Is it then up to the developer to then code a fall-back mechanism for such a case, otherwise their app could be in an invalid state.

Perhaps there should be a version of this method that does not time out? (effectively the opposite problem to `waitForChildAttached`)

core/src/main/java/com/bumble/appyx/core/node/ParentNode.kt
KovalevAndrey: Added a javadoc

zsoltk (approved):
Minor things:

1. Maybe better with naming like `ChildOne`, `GrandchildTwo`, etc.
2. The 5 second timeout seems too long. Maybe default 1s, and allow override if someone wants? We can get back to this later too, not blocking.

core/src/main/java/com/bumble/appyx/core/node/ParentNode.kt
zsoltk: I feel this error should definitely be instant. Even the 1 sec seems like a technicality. Increasing it further is bad ux (devx) imo. A random crash some time later, disconnected from the action itself. Can we instead get around it anyhow and get an instant crash?

sandbox/src/main/java/com/bumble/appyx/sandbox/client/workflow/WorkflowExampleActivity.kt
KovalevAndrey: decided to address this later

core/src/main/java/com/bumble/appyx/core/node/ParentNode.kt
KovalevAndrey: I made 1 sec by default with ability to override it. I don't like it as it think it shouldn't be part of the API but we can revise it later

PR #48, opened 2022-08-01
Converted customisations module from Android to Java

### LINKED ISSUE #47: Convert 'customisations' module from Android to Java
The 'customisations' does not need to be an Android module, and should be converted to a Java module. Should we wish to support multiplatform in the future we will need these modules to not be Android specific

LachlanMcKee:
## Description
Fixes #47

## Check list

### DIFF
--- a/gradle/publication.gradle
+++ b/gradle/publication.gradle
@@ -1,12 +1,19 @@
 apply plugin: 'maven-publish'
 apply plugin: 'signing'
 
-android {
-    publishing {
-        singleVariant('release') {
-            withSourcesJar()
-            withJavadocJar()
+if (pluginManager.hasPlugin("com.android.library")) {
+    android {
+        publishing {
+            singleVariant('release') {
+                withSourcesJar()
+                withJavadocJar()
+            }
         }
+    }
+} else {
+    java {
+        withJavadocJar()
+        withSourcesJar()
     }
 }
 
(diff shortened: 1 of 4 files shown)


gradle/publication.gradle
```diff
@@ -1,20 +1,31 @@
...
+    android {
+        publishing {
+            singleVariant('release') {
+                withSourcesJar()
+                withJavadocJar()
+            }
         }
     }
+} else {
+    java {
```
LachlanMcKee, line 14: I've verified this works by using `publishToMavenLocal` and inspecting the output

gradle/publication.gradle
```diff
@@ -1,20 +1,31 @@
 apply plugin: 'maven-publish'
 apply plugin: 'signing'
 
-android {
-    publishing {
-        singleVariant('release') {
```
CherryPerry, line 6: Can you name it `'java'`, so we do not need to add another split in `publishing` block?

LachlanMcKee: @CherryPerry I don't believe so. This is the `release` variant of the library as opposed to `debug`. If I named it `java` it wouldn't find the build.

Also, I am improving the plugin in this PR: https://github.com/bumble-tech/appyx/pull/51

CherryPerry:
I hope it won't break transitive dependency in RIBs.

LachlanMcKee:
If this does cause an issue with RIBs, I will revert this. I am pretty confident it should not cause an issue however

PR #51, opened 2022-08-01
Created publish plugin

### LINKED ISSUE #50: Create publishing plugin
Rather than using the existing `publication.gradle` class to setup our publication, we should use a plugin instead as this will give us greater type safety when switching to Kotlin gradle files (see #49)

LachlanMcKee:
## Description
Fixes #50.

Note: This relies on #48 being merged first

## Check list


build.gradle
```diff
@@ -22,3 +22,12 @@ allprojects {
         }
     }
 }
+
+// Required for the 'publish' plugin
+subprojects {
```
CherryPerry, line 27: Add TODO here, `subprojects` block should not be used to declare repositories.

gradle.properties
```diff
@@ -3,3 +3,4 @@ org.gradle.caching=true
 org.gradle.parallel=true
 android.useAndroidX=true
 kotlin.code.style=official
+library.version=1.0-alpha02
```
CherryPerry, line 6: alpha03

plugins/publish-plugin/build.gradle.kts
```diff
@@ -0,0 +1,34 @@
...
+  repositories {
+    google()
+    mavenCentral()
+  }
+}
+
+plugins {
+  `java-gradle-plugin`
+  `kotlin-dsl`
+  `kotlin-dsl-precompiled-script-plugins`
```
CherryPerry, line 15: Do we need it? AFAIK it is for `gradle.kts` files that you want to put into `src` folder.

gradle.properties
LachlanMcKee: done

plugins/publish-plugin/build.gradle.kts
LachlanMcKee: removed

build.gradle
LachlanMcKee: removed. I added a `pluginManagement` to the main `settings.gradle`

settings.gradle
```diff
@@ -1,3 +1,10 @@
+pluginManagement {
+    repositories {
+        google()
```
LachlanMcKee, line 3: required, otherwise the plugin is unable to locate `com.android.tools.build:gradle:7.1.3`, it tries to find it within the gradle plugin portal only

LachlanMcKee:
@CherryPerry I have addressed your comments :)

CherryPerry (approved):

PR #57, opened 2022-08-05
Moved appcompat classes out of core

### LINKED ISSUE #56: Move appcompat code out of core module
The `core` module currently depends on the Android appcompat library using `implementation`. Rather than exposing it as `api` it might be worthwhile to extract this into `core-appcompat`, as this will give developers a choice (as they may not be allowed to use the appcompat library)

LachlanMcKee:
## Description
Fixes #56 

## Check list


LachlanMcKee:
Closing for now as we still need to make a decision as to whether we want this extra module or not.

PR #61, opened 2022-08-11
[bumble-tech#59] Added interface for ParentNodeView

sfeatherstone:
## Description

Change `ParentNodeView` to inherit from a interface as described in issue [bumble-tech#59]

## Check list


core/src/main/java/com/bumble/appyx/core/node/ParentNodeViewBase.kt
```diff
@@ -0,0 +1,9 @@
+package com.bumble.appyx.core.node
+
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import com.bumble.appyx.core.plugin.NodeAware
+
+interface ParentNodeViewBase<Routing : Any> : NodeView, NodeAware<ParentNode<Routing>> {
```
LachlanMcKee, line 7: While this would be a breaking change, I personally would prefer if we renamed the abstract class, and give this interface the non base name

sfeatherstone: It would mean minimal changes to my current project so I have no opinion each way.
Please suggest the name for the implementation if it goes that way

mapm14 (approved):
LGTM :)

core/src/main/java/com/bumble/appyx/core/node/ParentNodeView.kt
```diff
@@ -4,7 +4,7 @@ import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
 import com.bumble.appyx.core.plugin.NodeAware
 
-abstract class ParentNodeView<Routing : Any> : NodeView, NodeAware<ParentNode<Routing>> {
+abstract class ParentNodeView<Routing : Any> : ParentNodeViewBase<Routing> {
```
KovalevAndrey, line 7: Can't say I'm a fan of the `Abstract` prefix but I don't have other options
```suggestion
abstract class AbstractParentNodeView<Routing : Any> : ParentNodeView<Routing> {
```

core/src/main/java/com/bumble/appyx/core/node/ParentNodeViewBase.kt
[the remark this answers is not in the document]
KovalevAndrey: I think we should leave interface type in the `ParentNode`

core/src/main/java/com/bumble/appyx/core/node/ParentNodeView.kt
```diff
@@ -4,21 +4,6 @@ import androidx.compose.runtime.Composable
...
-    @Composable
-    override fun View(modifier: Modifier) {
-        node.NodeView(modifier = modifier)
-    }
-
-    @Composable
-    abstract fun ParentNode<Routing>.NodeView(modifier: Modifier)
-
+interface ParentNodeView<Routing : Any> : NodeView, NodeAware<ParentNode<Routing>> {
+    @Composable fun ParentNode<Routing>.NodeView(modifier: Modifier)
```
KovalevAndrey, line 8:
```suggestion
    
    @Composable
    fun ParentNode<Routing>.NodeView(modifier: Modifier)
```

KovalevAndrey (approved):

PR #63, opened 2022-08-11
Report Android Lint to CodeQL

CherryPerry:
## Description

Display Android Lint errors directly in PR.
Setup detekt later.
Enable strict lint later (remove `abortOnError false`).

## Check list


plugins/sarif-plugin/src/main/kotlin/ReportLintSarifPlugin.kt
```diff
@@ -0,0 +1,31 @@
+import com.android.build.api.dsl.LibraryExtension
+import com.android.build.gradle.internal.lint.AndroidLintTask
+import org.gradle.api.Plugin
+import org.gradle.api.Project
+import org.gradle.kotlin.dsl.configure
+
+class ReportLintSarifPlugin : Plugin<Project> {
+
+    override fun apply(target: Project) {
+        target.plugins.withId("com.android.library") {
```
LachlanMcKee, line 10: What about the app/sandbox modules?

plugins/sarif-plugin/src/main/kotlin/ReportLintSarifPlugin.kt
```diff
@@ -0,0 +1,31 @@
...
+            target.extensions.configure<LibraryExtension> {
+                lint.sarifReport = true
+            }
+            val rootProject = target.rootProject
+            rootProject.plugins.withId("appyx-collect-lint-sarif") {
```
LachlanMcKee, line 15: If this plugin is not found, is there any point in setting `lint.sarifReport = true`?

plugins/sarif-plugin/src/main/kotlin/ReportLintSarifPlugin.kt
CherryPerry: IMHO do not see any value in linting these modules.

PR #65, opened 2022-08-12
Add interop builder stub

jlmd:
## Description
Add `InteropBuilderStub` and `InteropSimpleBuilderStub` util classes to allow testing that a child node of appyx has been created within a RIBs parent.

## Check list


sandbox/src/main/java/com/bumble/appyx/sandbox/client/interop/parent/routing/RibsParentChildBuilders.kt
```diff
@@ -0,0 +1,18 @@
+package com.bumble.appyx.sandbox.client.interop.parent.routing
+
+import com.badoo.ribs.builder.Builder
+import com.bumble.appyx.core.builder.SimpleBuilder
+import com.bumble.appyx.sandbox.client.container.ContainerBuilder
+import com.bumble.appyx.sandbox.client.interop.child.RibsChild
+import com.bumble.appyx.sandbox.client.interop.child.RibsChildBuilder
+
+interface RibsParentChildBuilders {
```
mapm14, line 9:
```suggestion
internal interface RibsParentChildBuilders {
```

mapm14 (approved):
Great job, Jose! 👏🏻

sandbox/src/main/java/com/bumble/appyx/sandbox/client/interop/child/RibsChild.kt
```diff
@@ -1,11 +1,14 @@
 package com.bumble.appyx.sandbox.client.interop.child
 
 import com.badoo.ribs.core.Node
+import com.badoo.ribs.core.Rib
 import com.badoo.ribs.core.modality.BuildParams
 import com.badoo.ribs.core.view.RibView
 
+interface RibsChild : Rib
```
zsoltk, line 8: Is this needed?

jlmd: I added it to be able to create the `RibsChildBuilder`, so both children are created using builders and also to create [this test](https://github.com/bumble-tech/appyx/pull/65/files#diff-7a3789fc1583c7c46c8dfaa87fedab0b959dc4fea991639161f679c206e1f441R72) as an example

testing-unit-common/src/main/java/com/bumble/appyx/testing/unit/common/util/InteropSimpleBuilderStub.kt
```diff
@@ -0,0 +1,31 @@
...
+        delegate(buildContext).also {
+            _createdNodes += it
+        }
+
+    fun assertCreatedNode() {
+        assertTrue(last != null, "Has not created any node")
+    }
+
+    fun assertLastNodeState(state: Lifecycle.State) {
+        assertEquals(state, last?.lifecycle?.currentState)
```
LachlanMcKee, line 29: for now if `last?.lifecycle?.currentState` returns null, you know that the last node was null. however potentially the API could change in the future which may make this test have a false positive/negative (i.e. how lifecycle is nullable, but the last node was not null, making it confusing to tell why the test failed, or worse it passes).

I proposed the following (because we don't seem to have soft asserts):
```
assertCreatedNode()
assertEquals(state, last!!.lifecycle.currentState)
```

This way if the API of lifecycle changes to be nullable, the code would not compile.

testing-unit-common/src/main/java/com/bumble/appyx/testing/unit/common/util/InteropSimpleBuilderStub.kt
```diff
@@ -0,0 +1,31 @@
...
+import com.bumble.appyx.core.modality.BuildContext
+import com.bumble.appyx.core.node.Node
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+
+class InteropSimpleBuilderStub(
+    val delegate: (BuildContext) -> Node = { _ -> NodeStub() },
+) : SimpleBuilder() {
+
+    private val _createdNodes = ArrayList<Node>()
```
LachlanMcKee, line 14: At the moment does this need to be a list?

testing-unit-common/src/main/java/com/bumble/appyx/testing/unit/common/util/InteropBuilderStub.kt
```diff
@@ -0,0 +1,46 @@
...
+            _createdNodes += it
+            lastParam = payload
+        }
+
+    fun assertLastParam(param: P) {
+        assertEquals(param, lastParam)
+    }
+
+    fun assertLastParam(assert: P.() -> Boolean) {
+        assertTrue(
```
LachlanMcKee, line 33: I would split these two assertions into two, otherwise it's tricky to tell what the problem is.

testing-unit-common/src/main/java/com/bumble/appyx/testing/unit/common/util/InteropBuilderStub.kt
```diff
@@ -0,0 +1,46 @@
...
+            "$lastParam does not satisfy requirements"
+        )
+    }
+
+    fun assertCreatedNode() {
+        assertTrue(last != null, "Has not created any node")
+    }
+
+    fun assertLastNodeState(state: Lifecycle.State) {
+        assertEquals(state, last?.lifecycle?.currentState)
```
LachlanMcKee, line 44: for now if last?.lifecycle?.currentState returns null, you know that the last node was null. however potentially the API could change in the future which may make this test have a false positive/negative (i.e. how lifecycle is nullable, but the last node was not null, making it confusing to tell why the test failed, or worse it passes).

I proposed the following (because we don't seem to have soft asserts):

assertCreatedNode()
assertEquals(state, last!!.lifecycle.currentState)
This way if the API of lifecycle changes to be nullable, the code would not compile.

sandbox/src/main/java/com/bumble/appyx/sandbox/client/interop/parent/RibsParentRib.kt
```diff
@@ -2,18 +2,18 @@ package com.bumble.appyx.sandbox.client.interop.parent
...
+import com.bumble.appyx.utils.customisations.NodeCustomisation
 
 interface RibsParentRib : Rib {
 
     class Customisation(
-        val viewFactory: Factory = Factory(),
+        val viewFactory: Factory = RibsParentViewImpl.Factory(),
         val name: String = "defaultName"
-    ) : RibCustomisation
+    ) : NodeCustomisation
```
LachlanMcKee, line 16: Is this still technically completely testing interop of a pure RIB? you've leaked appyx into this class

testing-unit-common/src/main/java/com/bumble/appyx/testing/unit/common/util/InteropBuilderStub.kt
jlmd: I don't think it's necessary to split this 🤔 
If the test fails it will tell you the `lastParams` value in the message, so you will see if it failed because it's null or because it's different from the expected assert

testing-unit-common/src/main/java/com/bumble/appyx/testing/unit/common/util/InteropSimpleBuilderStub.kt
jlmd: I like it, thanks!
Changed in 592c9d9268cfd05127ea0c4739787a7430fd4fcc and 8cad9d2c97c70f01fe62e2a4b1ed603319dcd022

testing-unit-common/src/main/java/com/bumble/appyx/testing/unit/common/util/InteropSimpleBuilderStub.kt
jlmd: Not really, I changed it in 592c9d9268cfd05127ea0c4739787a7430fd4fcc

sandbox/src/main/java/com/bumble/appyx/sandbox/client/interop/parent/RibsParentRib.kt
jlmd: You are right! I reverted it in 592c9d9268cfd05127ea0c4739787a7430fd4fcc

testing-unit-common/src/main/java/com/bumble/appyx/testing/unit/common/util/InteropBuilderStub.kt
LachlanMcKee: the key point is that there is an 'or' involved here which adds ambiguity . Splitting it doesn't cost us much, and can make the users life easier.

They can see that 'oh last param is null' or the param didn't pass the predicate.

jlmd: Done in 9733f65c631ba6a3d30af59207cebeb095f1e872

KovalevAndrey (approved):

sandbox/src/main/java/com/bumble/appyx/sandbox/client/interop/child/RibsChild.kt
jlmd: It was not needed, I removed it in 4cec1addcaa4010a67d8fe2b6b262e947952fc5b and used `Rib` directly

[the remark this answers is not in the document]
jlmd: You are right! I removed it in 4cec1addcaa4010a67d8fe2b6b262e947952fc5b

PR #67, opened 2022-08-12
Lint & detekt with GitHub reports

CherryPerry:
## Description

- Lint setup with baselines
- Detekt setup with baselines
- Generate SARIF file with both tools and merge them
- New verification plugin module to make it work
- Gradle upgrade as part of bug investigation

I am not sure how stable does it work (regarding posting comments with issues), but I can fix it on the fly. Let's see how it works. Also I am not sure about detekt errors, I tried to increase their level in `CollectSarifPlugin.enforceErrorLevel` but looks like it does not work, need to check later.

Screenshot 2022-08-12 at 15 56 32

## Check list


app/detekt-baseline.xml
```diff
@@ -0,0 +1,361 @@
+<?xml version='1.0' encoding='UTF-8'?>
```
LachlanMcKee, line 1: I assume we will fix some of these in the future? (same question for the other modules)

LachlanMcKee (approved):

app/detekt-baseline.xml
CherryPerry: I hope so.
At least I do not want to do it right now, as it will take a lot of time.

PR #72, opened 2022-08-13
Fixed existing detekt/lint issues

LachlanMcKee:
## Description
Fixed the majority of the outstanding detekt/lint errors.
The remaining issues should be investigated further.

## Check list


testing-ui/src/main/java/com/bumble/appyx/testing/ui/rules/AppyxTestRule.kt
```diff
@@ -16,6 +16,7 @@ open class AppyxTestRule<T : Node>(
     private val launchActivity: Boolean = true,
     private val composeTestRule: ComposeTestRule = createEmptyComposeRule(),
     /** Add decorations like custom theme or CompositionLocalProvider. Do not forget to invoke `content()`. */
+    @Suppress("unused")
```
LachlanMcKee, line 19: unsure about whether to remove this variable or not

CherryPerry: Keep it and create an issue, I forgot to use it.

sandbox/src/main/res/values/strings.xml
```diff
@@ -1,3 +1,6 @@
 <resources>
     <string name="app_name">Appyx sandbox</string>
+    <string name="rib_root_title">Black area belongs to 1.0, container below to 2.0</string>
```
CherryPerry, line 3: I think you could just add `HardcodedText` to Lint ignore list for sandbox/app projects.

plugins/verification-plugin/src/main/kotlin/DetektPlugin.kt
```diff
@@ -13,7 +13,14 @@ class DetektPlugin : Plugin<Project> {
             target.extensions.configure<DetektExtension> {
                 buildUponDefaultConfig = true
                 baseline = target.file("detekt-baseline.xml")
-                config.from(target.rootProject.file("detekt.yml"))
+
+                val localDetektConfig = target.file("detekt.yml")
+                val rootDetektConfig = target.rootProject.file("detekt.yml")
```
CherryPerry, line 18: Does specifying both files always instead of checking existence work?

CherryPerry (approved):

sandbox/src/main/res/values/strings.xml
LachlanMcKee: done

plugins/verification-plugin/src/main/kotlin/DetektPlugin.kt
LachlanMcKee: looks like that works. good idea!

LachlanMcKee: scratch that... I didn't run it correctly locally when testing. reverted back to how it was.

Issue #81, opened 2022-08-16
Support integration point for multiple roots


KovalevAndrey:
At the moment `IntegrationPoint` is designed to have only one root in Appyx tree and will crash if we try to attach another one. In reality there can be more than one Appyx tree (similar to composition). For instance, clients may want to use Appyx in their codebase in disconnected from each other places creating more than one tree

PR #82, opened 2022-08-16
Remove assumption that there can be only one root from IntegrationPoint

KovalevAndrey:
## Description

Fixes https://github.com/bumble-tech/appyx/issues/81

1. Removed assumption that only one Appyx root can exist
2. IntegrationPoint should be attached at the moment of the node creation. This way we can guarantee that a `Node` has `IntegrationPoint` before `node.Compose()` is called

## Check list

### DIFF
--- a/core/src/main/java/com/bumble/appyx/core/integration/NodeHost.kt
+++ b/core/src/main/java/com/bumble/appyx/core/integration/NodeHost.kt
@@ -2,6 +2,7 @@ package com.bumble.appyx.core.integration
 
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.DisposableEffect
+import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.State
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
@@ -29,9 +30,8 @@ fun <N : Node> NodeHost(
     factory: NodeFactory<N>
 ) {
     val node by rememberNode(factory, customisations)
-    DisposableEffect(integrationPoint) {
-        integrationPoint.attach(node)
-        onDispose { integrationPoint.detach() }
+    LaunchedEffect(integrationPoint, node) {
+        node.integrationPoint = integrationPoint
     }
     DisposableEffect(node) {
         onDispose { node.updateLifecycleState(Lifecycle.State.DESTROYED) }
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/bumble/appyx/core/integration/NodeHost.kt
```diff
@@ -29,9 +30,8 @@ fun <N : Node> NodeHost(
     factory: NodeFactory<N>
 ) {
     val node by rememberNode(factory, customisations)
-    DisposableEffect(integrationPoint) {
-        integrationPoint.attach(node)
-        onDispose { integrationPoint.detach() }
+    LaunchedEffect(integrationPoint, node) {
+        node.integrationPoint = integrationPoint
```
LachlanMcKee, line 34: Does this need to be in a launched effect?

KovalevAndrey: yes, because otherwise it will be called on every recomposition

LachlanMcKee: Good point

LachlanMcKee (approved):

core/src/main/java/com/bumble/appyx/core/integration/NodeHost.kt
```diff
@@ -28,11 +29,7 @@ fun <N : Node> NodeHost(
     integrationPoint: IntegrationPoint,
     factory: NodeFactory<N>
 ) {
-    val node by rememberNode(factory, customisations)
-    DisposableEffect(integrationPoint) {
-        integrationPoint.attach(node)
-        onDispose { integrationPoint.detach() }
-    }
+    val node by rememberNode(factory, customisations, integrationPoint)
```
CherryPerry, line 32: `NodeCustomisationDirectory` is neither Stable nor Immutable, does it affects us?
Also if I do not pass `customisations` we create new `NodeCustomisationDirectoryImpl` on each composition.
(can be addressed as a separate PR)

CherryPerry (approved):

core/src/main/java/com/bumble/appyx/core/integration/NodeHost.kt
KovalevAndrey: good point, we can remember it to prevent unnecessary recompositions

PR #95, opened 2022-08-22
Add interactor tests when parent node implements Connectable

mapm14:
## Description
This PR fix a crash (on unit tests) when using an `InteractorTestHelper`, the steps to reproduce it are:

- Have a Node that implements `Connectable<Input, Output>` 
- In the Interactor of that Node we add a binding that uses either the `node.input` or `node.output`
- Create an Interactor unit test that uses the current `InteractorTestHelper`

Crashlogs:
image

I'm not sure if this is the best approach and there are some classes name I wasn't sure about it but probably we can fix that more easily :)
...

## Check list


CherryPerry (changes requested):

CherryPerry:
https://github.com/bumble-tech/appyx/pull/96

PR #96, opened 2022-08-22
Remove InteractorTestHelper.kt

mapm14:
## Description
We're discouraging the use of Interactor tests in favor of using Node tests.

The reasoning here is that unit tests are about units, not classes. And in the case of a node-based architecture, the unit is the node. Also by writing node tests you are also covering the integration of different node parts.

## Check list


LachlanMcKee:
@mapm14 can you please update the changelog?

LachlanMcKee (approved):

CherryPerry (approved):
Approved with [the issue](https://github.com/bumble-tech/appyx/issues/101).

`MviCoreExampleNode` should not accept node connectors for their children, it looks strange.

KovalevAndrey:
> Approved with [the issue](https://github.com/bumble-tech/appyx/issues/101).
> 
> `MviCoreExampleNode` should not accept node connectors for their children, it looks strange.

it also makes MVI Core example crash

PR #104, opened 2022-08-25
[#53] (Proposal 1) Removed IntegrationPointAppyxProvider

### LINKED ISSUE #53: Interop Integrationpoint shouldn't force to expose appyx interface
Current interop integration point required activities to implement `IntegrationPointAppyxProvider`.

This expose the interface in module edges, forcing parent modules to implement appyx event if they are not using it.

Example.

`Module A` -> `Module B`

`Module B` implements `SomeActiviy: RibAcitivty(), IntegrationPointAppyxProvider{...}`
`Module A` has no appyx dependencies.

Compilation failed as Module A will try to resolve  `IntegrationPointAppyxProvider`

LachlanMcKee:
## Description
Possible solution for avoiding exposing the appyx dependencies in modules that reference an Activity that extends IntegrationPointAppyxProvider

Fixes #53 

## Check list


core/src/main/kotlin/com/bumble/appyx/core/integrationpoint/ActivityIntegrationPoint.kt
```diff
@@ -44,4 +48,44 @@ open class ActivityIntegrationPoint(
...
     }
+
+    companion object {
+        private val integrationPoints = WeakHashMap<AppCompatActivity, IntegrationPoint>()
+
+        fun createIntegrationPoint(
+            activity: AppCompatActivity,
+            savedInstanceState: Bundle?,
+        ): ActivityIntegrationPoint =
+            ActivityIntegrationPoint(activity, savedInstanceState)
```
KovalevAndrey, line 59: shouldn't we check for existing integration point for this activity before creating it?

KovalevAndrey (approved):

core/src/main/kotlin/com/bumble/appyx/core/integrationpoint/ActivityIntegrationPoint.kt
```diff
@@ -44,4 +48,44 @@ open class ActivityIntegrationPoint(
             activity.finish()
         }
     }
+
+    companion object {
+        private val integrationPoints = WeakHashMap<AppCompatActivity, IntegrationPoint>()
+
+        fun createIntegrationPoint(
+            activity: AppCompatActivity,
```
KovalevAndrey, line 56: let's align with https://github.com/bumble-tech/appyx/pull/132/files#diff-61d49596405e9cbcac4b41945813a8097a38cb5adecb8b0826061acc93fc3fe4R12

LachlanMcKee: @KovalevAndrey done

core/src/main/kotlin/com/bumble/appyx/core/integrationpoint/ActivityIntegrationPoint.kt
LachlanMcKee: The existing API doesn't work this way. We could do a getOrCreate or something similar, but I wanted to remain reasonably consistent.

KovalevAndrey (approved):

PR #115, opened 2022-08-27
[#144] Add dagger hilt compiler and sample project

### LINKED ISSUE #114: Consider adding dagger hilt integration
We could introduce some simple dagger hilt integration to make it easier to share node factories across modules without exposing internal dependencies.

LachlanMcKee:
## Description
Fixes #114 

## Check list
(will update depending on the outcome of the proposal)


gradle/libs.versions.toml
```diff
@@ -1,7 +1,7 @@
 # keep sorted please, use Edit -> Sort Lines
 
 [versions]
-androidx-lifecycle = "2.4.1"
+androidx-lifecycle = "2.5.1"
```
LachlanMcKee, line 4: I will fix this file depending on what the outcome of this proposal is.
We should update these dependencies in another PR. Without updating them, hilt causes issues.

dagger/hilt-node-factory-runtime/src/main/kotlin/com/bumble/appyx/dagger/hilt/node/AppyxNodeFactory.kt
```diff
@@ -0,0 +1,3 @@
+package com.bumble.appyx.dagger.hilt.node
+
+interface AppyxNodeFactory
```
LachlanMcKee, line 3: This may not be required, though without it, there effectively no type safety.

dagger/hilt-node-factory-compiler/src/main/kotlin/com/bumble/appyx/dagger/hilt/node/AppyxNodeFactoryProcessor.kt
```diff
@@ -0,0 +1,127 @@
...
+            roundEnv.getElementsAnnotatedWith(processingAnnotation).forEach {
+                processElement(it as TypeElement)
+            }
+        }
+        return false
+    }
+
+    private fun processElement(typeElement: TypeElement) {
+        val bindingClassName = ClassName.get(typeElement)
+        val superTypeClassName = ClassName.get(typeElement.interfaces[0])
```
LachlanMcKee, line 55: This is very very brittle. I will make finding the interface safer if we want to go ahead with this change

core/src/test/kotlin/com/bumble/appyx/core/lifecycle/MinimumCombinedLifecycleTest.kt
```diff
@@ -40,9 +40,13 @@ class MinimumCombinedLifecycleTest {
     @Test
     fun `destroyed when one of lifecycles is destroyed`() {
         val test1 = TestLifecycle()
-        test1.state = State.DESTROYED
+        test1.state = State.CREATED
```
LachlanMcKee, line 43: This is due to updating the Lifecycle library (unrelated to this change). I'm unsure if this is the correct way to fix this, but it wouldn't be merged in this PR (it would be in a different PR).

Without this change, the test fails with: `no event down from INITIALIZED` because they don't want you to immediate move to the DESTROYED state.

dagger/hilt-node-factory-runtime/src/main/kotlin/com/bumble/appyx/dagger/hilt/node/AppyxNodeFactories.kt
```diff
@@ -0,0 +1,46 @@
...
+class AppyxNodeFactories(
+    private val delegate: Map<Class<*>, Provider<AppyxNodeFactory>>
+) : Map<Class<*>, Provider<AppyxNodeFactory>> by delegate
+
+inline fun <reified T : AppyxNodeFactory> AppyxNodeFactories.getAppyxNodeFactory(): T {
+    return (getValue(T::class.java) as Provider<T>).get()
+}
+
+fun appyxNodeFactories(context: Context): AppyxNodeFactories {
+    val activity = context.let {
```
LachlanMcKee, line 27: This activity finding logic could be shared with the ribs-interop module through a common library or part of core

samples/dagger-hilt/library/src/main/kotlin/com/bumble/appyx/sample/hilt/library/InternalDependencyLibraryModule.kt
```diff
@@ -0,0 +1,15 @@
...
+import dagger.Module
+import dagger.Provides
+import dagger.hilt.InstallIn
+import dagger.hilt.android.components.ActivityComponent
+
+@Module
+@InstallIn(ActivityComponent::class)
+internal object InternalDependencyLibraryModule {
+    @Provides
+    fun provideExternalDepsDependency(activity: Activity): InternalDependency =
```
Ninjars: minor typo? "provideExternalDepsDependency" when everything else is InternalDependency? Or is that referencing that the internal dependency is a dependency for the external dependency...?

samples/dagger-hilt/library/src/main/kotlin/com/bumble/appyx/sample/hilt/library/InternalDependencyLibraryModule.kt
```diff
@@ -0,0 +1,15 @@
...
+
+import android.app.Activity
+import dagger.Module
+import dagger.Provides
+import dagger.hilt.InstallIn
+import dagger.hilt.android.components.ActivityComponent
+
+@Module
+@InstallIn(ActivityComponent::class)
+internal object InternalDependencyLibraryModule {
```
Ninjars, line 11: is this module for demo purposes, or is this required? I would have expected to be just able to use `@Inject constructor()` in `InternalDependency`, but wondering if there was something I was missing.

samples/dagger-hilt/library/src/main/kotlin/com/bumble/appyx/sample/hilt/library/externaldeps/ExternalDepsLibraryCustomNodeFactory.kt
```diff
@@ -0,0 +1,39 @@
...
+    override fun create(
+        buildContext: BuildContext,
+        hiltNodeFactoryProvider: HiltNodeFactoryProvider
+    ) =
+        ExternalDepsLibraryNodeImpl(
+            buildContext = buildContext,
+            noDepsLibraryNodeFactory = hiltNodeFactoryProvider.getNodeFactory(),
+            dependency = internalDependency
+        )
+}
```
Ninjars: I'm not sure I'm following what this is doing/what the purpose is. Am I right in thinking that this is a factory that you use to build a node that receives dependencies provided by hilt?

So when you have a navigation destination, instead of creating the node directly you have a factory that builds it?

Does that mean that for every node you need injection into you also need to define a factory? My initial reaction to the shape of that idea is negative; part of the point of Hilt is to reduce boiler plate and factories calling factories feels antithetical to that.

I guess my "more idea" naive expectation would be something like how the `hiltViewModel()`convenience function works with the AAC ViewModel: the function infers from the type what to construct and provides the dependencies from the graph automatically. I assume it generates a factory for that behind the scenes.

Though it is worth noting that that function falls down when you want to be able to provide parameters and you're back to defining a bunch of boiler plate yourself, if memory serves.

LachlanMcKee: You don't need to provide a factory for every node, just nodes that take dependencies. Take a look at the `NoDepsLibraryNode` for example.

You also don't need to pass `hiltNodeFactoryProvider` to all nodes, just nodes that might want to create a node using this approach, it's optional and just for demonstration purposes.

Ninjars: so if you were using Appyx as the core navigational framework for your whole application you'd be needing dependencies pretty commonly. To satisfy that you'd need a lot of factories to propagate the hiltNodeFactoryProvider throughout the nav graph, right?

I'm just wondering what this would look like in practice in a non-trivial app or at-scale in general and wondering if there's something we can do to avoid boilerplate dependency provisioning in the common usecase of providing a set of dependencies to a node that may be nested arbitrarily deep in a navigation graph.

LachlanMcKee:
Ideas based on discussion:
- Have examples where the nodes are not internal, see how much simpler we can make the implementation for users.

PR #116, opened 2022-08-28
Moved compile/target/min sdks into version catalog

LachlanMcKee:
## Description
Reduces repetition and allows us to change the compile/target/min sdks in one place

## Check list


samples/navigation-compose/build.gradle.kts
```diff
@@ -0,0 +1,57 @@
+plugins {
```
LachlanMcKee, line 1: This file missed the migration (probably due to multiple PRs open), hopefully it's fine to fix this in this PR.

gradle/libs.versions.toml
```diff
@@ -1,6 +1,9 @@
 # keep sorted please, use Edit -> Sort Lines
 
 [versions]
+androidCompileSdk = "32"
+androidMinSdk = "21"
+androidTargetSdk = "32"
```
CherryPerry, line 6: You can use the same version for `compile` and `target`, I do not think we will have time when these 2 will differ.

CherryPerry (approved):

gradle/libs.versions.toml
LachlanMcKee: Hopefully we won't, but at least it's centralised here, and if we do need to for some reason, we have the flexibility to do so.

Issue #120, opened 2022-08-31
BuildContext.savedStateMap memory leak


CherryPerry:
`BuildContext.savedStateMap` is not cleared after restoration, effectively holding up to 1 MB of data until the next restoration cycle.

Issue #111, opened 2022-08-26
Make logic of keeping off screen Nodes alive configurable


KovalevAndrey:
Keeping all Nodes in the NavModel alive while they're invisible can be memory consuming. We need to make it configurable. Twitter thread for the reference: https://twitter.com/Piwai/status/1563157989430661122?s=20&t=FoQaDoEPift8J2TGkqOuGQ

saket:
+1 for this. I ran into this yesterday when I was trying to only keep fullscreen nodes in memory that have a transparent background.

PR #121, opened 2022-08-31
Extract child management logic from ParentNode

CherryPerry:
## Description

Extract the logic to make it easier to maintain later for https://github.com/bumble-tech/appyx/issues/111
Fixes partially https://github.com/bumble-tech/appyx/issues/120 (have not checked if completely).

Not a behavioural change - do not mention it in the change log.

## Check list

### DIFF
--- a/core/src/main/kotlin/com/bumble/appyx/core/children/ChildNodeCreationManager.kt
+++ b/core/src/main/kotlin/com/bumble/appyx/core/children/ChildNodeCreationManager.kt
@@ -1,0 +1,112 @@
+package com.bumble.appyx.core.children
+
+import androidx.lifecycle.coroutineScope
+import com.bumble.appyx.core.modality.AncestryInfo
+import com.bumble.appyx.core.modality.BuildContext
+import com.bumble.appyx.core.navigation.RoutingKey
+import com.bumble.appyx.core.node.ParentNode
+import com.bumble.appyx.core.state.MutableSavedStateMap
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/com/bumble/appyx/core/children/ChildNodeCreationManager.kt
```diff
@@ -0,0 +1,112 @@
...
+                    when (updateChild) {
+                        is ChildEntry.Eager -> map
+                        is ChildEntry.Lazy -> map.plus(routingKey to updateChild.initialize())
+                    }
+                }[routingKey] as ChildEntry.Eager
+        }
+    }
+
+    private fun SavedStateMap?.restoreChildren(parentNode: ParentNode<Routing>): ChildEntryMap<Routing>? =
+        (this?.get(KEY_CHILDREN_STATE) as? Map<RoutingKey<Routing>, SavedStateMap>)?.mapValues {
```
LachlanMcKee, line 80: Worth moving the private functions to the bottom? Also, can they be added to the companion object?

CherryPerry: It will become non-static in the next PR.

In general Kotlin code style recommends to group methods by context and not visibility, but I do not have any strong opinion on it.

LachlanMcKee (approved):

PR #122, opened 2022-08-31
Suspend children

CherryPerry:
## Description

Simple implementation with either keep all or suspend all.

1. New `KeepMode` is introduced to specify keep logic of stashed (not visible) nodes. The value is passed directly into `ParentNode`.
2. `ChildEntry` is refactored into `Initialized` and `Suspended`.
3. `ChildNodeCreationManager` creates a snapshot of currently initialized and suspended nodes and compares it with the provided by nav model list. After calculating a diff, it applies required changes.
4. Simple tests to verify behaviour.
5. `Appyx` class moved to the root package and allows to setup default behaviour.

Fixes https://github.com/bumble-tech/appyx/issues/111
Fixes https://github.com/bumble-tech/appyx/issues/120

### Strategies

As a separate PR please, this one defines a foundation for further updates and allows faster feedback.

## Check list


core/src/main/kotlin/com/bumble/appyx/core/children/ChildNodeCreationManager.kt
```diff
@@ -0,0 +1,178 @@
...
+        }
+        parentNode.lifecycle.coroutineScope.launch {
+            // TODO Extract as strategy? Then easy to miss bug fixes to both.
+            parentNode.navModel.screenState.collect { state ->
+                // TODO Do not calculate it if wrong keep mode
+                val navModelOnScreenKeys = state.onScreen
+                    .mapTo(HashSet(state.onScreen.size, 1f)) { element -> element.key }
+                val navModelOffScreenKeys = state.offScreen
+                    .mapTo(HashSet(state.offScreen.size, 1f)) { element -> element.key }
+                val navModelKeys = navModelOnScreenKeys + navModelOffScreenKeys
```
KovalevAndrey, line 44: I imagined it in a different way. Similarly to `onScreen/offScreen`, can't we expose `elementsToSave/elementsToRestore` and implement a strategy in `BaseNavModel` which will do the mapping of the elements with default strategy of mapping `onScreen/offScreen` to `elementsToRestore/elementsToSave`?

core/src/main/kotlin/com/bumble/appyx/core/children/ChildNodeCreationManager.kt
```diff
@@ -0,0 +1,178 @@
...
+    private val customisations: NodeCustomisationDirectory,
+    private val childMode: ChildEntry.ChildMode,
+    private val keepMode: ChildEntry.KeepMode,
+) {
+    private val _children =
+        MutableStateFlow<Map<RoutingKey<Routing>, ChildEntry<Routing>>>(emptyMap())
+    val children: StateFlow<ChildEntryMap<Routing>> = _children.asStateFlow()
+    private lateinit var parentNode: ParentNode<Routing>
+
+    fun launch(parentNode: ParentNode<Routing>) {
```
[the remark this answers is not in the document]
CherryPerry: Warning: LeakingThis (parent node is not final, also easy to make a mistake and move launch() into init that will cause a crash)

core/src/main/kotlin/com/bumble/appyx/core/children/ChildNodeCreationManager.kt
KovalevAndrey: I can also imagine that any particular Node can also have a flag to reflect the information if it needs to stay in the memory or not

core/src/main/kotlin/com/bumble/appyx/core/children/ChildNodeCreationManager.kt
```diff
@@ -15,88 +16,141 @@ import kotlinx.coroutines.flow.update
...
-                    val mutableMap = map.toMutableMap()
-                    newKeys.forEach { key ->
-                        mutableMap[key] =
-                            ChildEntry.create(
-                                key = key,
-                                resolver = parentNode,
-                                buildContext = null.toBuildContext(parentNode),
-                                childMode = childMode,
-                            )
+            parentNode.navModel.screenState.collect { state ->
```
KovalevAndrey, line 45: isn't it easier calculate the diff with `withPrevious()`?

CherryPerry: I do not need to compare with previous. I need to make sure that the nodes that I have in `children` are in the correct state (or removed). So I compare `navModel` with `children`.

core/src/main/kotlin/com/bumble/appyx/Appyx.kt
```diff
@@ -0,0 +1,19 @@
+package com.bumble.appyx
+
+import com.bumble.appyx.core.children.ChildEntry
+
+object Appyx {
+
+    var exceptionHandler: ((Exception) -> Unit)? = null
+    var defaultKeepMode: ChildEntry.KeepMode = ChildEntry.KeepMode.KEEP
```
KovalevAndrey, line 8: Make `Suspend` as default?

core/src/main/kotlin/com/bumble/appyx/core/children/ChildEntry.kt
```diff
@@ -20,45 +18,20 @@ sealed class ChildEntry<T> {
...
-            buildContext: BuildContext,
-            childMode: ChildMode,
-        ): ChildEntry<T> =
-            when (childMode) {
-                ChildMode.EAGER ->
-                    Eager(key, resolver.resolve(key.routing, buildContext).build())
-                ChildMode.LAZY ->
-                    Lazy(key, resolver, buildContext)
-            }
+    /** Keep not on screen nodes in the memory? */
```
KovalevAndrey, line 31: [nit] `Keep off screen nodes in the memory?`

KovalevAndrey (approved):
Please update `StatefulNode2`'s description. It currently says that Nodes are always alive. We can write something like "Node can be kept alive if you use KEEP mode"

core/src/main/kotlin/com/bumble/appyx/Appyx.kt
CherryPerry: The current default is `KEEP`, do not want to introduce breaking changes. I am pretty sure that we will find a lot of bugs because of it in our app.

KovalevAndrey: Yeah, you're right. We probably need to explicitly change mode to `Suspend` and analyse it in a migration task

core/src/main/kotlin/com/bumble/appyx/core/children/ChildEntry.kt
```diff
@@ -20,45 +18,20 @@ sealed class ChildEntry<T> {
...
     /** All public APIs should return this type of child which is ready to work with. */
-    class Eager<T>(
+    class Initialized<T>(
         override val key: RoutingKey<T>,
         val node: Node,
     ) : ChildEntry<T>()
 
-    /** Child representation for Lazy mode. */
-    class Lazy<T>(
+    class Suspended<T>(
```
LachlanMcKee, line 26: Add Javadoc to this class for consistency with `Initialized`?

core/src/main/kotlin/com/bumble/appyx/core/children/ChildEntry.kt
```diff
@@ -20,45 +18,20 @@ sealed class ChildEntry<T> {
...
-            buildContext: BuildContext,
-            childMode: ChildMode,
-        ): ChildEntry<T> =
-            when (childMode) {
-                ChildMode.EAGER ->
-                    Eager(key, resolver.resolve(key.routing, buildContext).build())
-                ChildMode.LAZY ->
-                    Lazy(key, resolver, buildContext)
-            }
+    /** Keep off-screen nodes in the memory? */
```
LachlanMcKee, line 31: Does this comment make sense for the enum as a whole, or just the KEEP enum?

core/src/main/kotlin/com/bumble/appyx/core/children/ChildNodeCreationManager.kt
```diff
@@ -15,88 +16,141 @@ import kotlinx.coroutines.flow.update
...
 internal class ChildNodeCreationManager<Routing : Any>(
     private var savedStateMap: SavedStateMap?,
     private val customisations: NodeCustomisationDirectory,
-    private val childMode: ChildEntry.ChildMode,
+    private val keepMode: ChildEntry.KeepMode,
 ) {
     private val _children =
         MutableStateFlow<Map<RoutingKey<Routing>, ChildEntry<Routing>>>(emptyMap())
     val children: StateFlow<ChildEntryMap<Routing>> = _children.asStateFlow()
+    private lateinit var parentNode: ParentNode<Routing>
```
LachlanMcKee, line 32: nit: group the private variables?

core/src/main/kotlin/com/bumble/appyx/debug/Appyx.kt
```diff
@@ -1,16 +1,9 @@
...
-            throw exception
-        }
-    }
-
-}
+@Deprecated(
+    message = "Moved to com.bumble.appyx package",
+    replaceWith = ReplaceWith("Appyx", imports = ["com.bumble.appyx.Appyx"]),
+)
+typealias Appyx = Appyx
```
LachlanMcKee, line 9: At what point do we remove this? 1.0.0? 1.1.0?

core/src/main/kotlin/com/bumble/appyx/core/node/ParentNode.kt
```diff
@@ -47,7 +48,7 @@ abstract class ParentNode<Routing : Any>(
     navModel: NavModel<Routing, *>,
     buildContext: BuildContext,
     view: ParentNodeView<Routing> = EmptyParentNodeView(),
-    childMode: ChildEntry.ChildMode = ChildEntry.ChildMode.EAGER,
+    keepMode: ChildEntry.KeepMode = Appyx.defaultKeepMode,
```
LachlanMcKee, line 51: `childKeepMode`? The name makes sense in `ChildNodeCreationManager`, but not necessarily for `ParentNode`

core/src/main/kotlin/com/bumble/appyx/core/children/ChildNodeCreationManager.kt
```diff
@@ -105,8 +159,59 @@ internal class ChildNodeCreationManager<Routing : Any>(
...
+        }
+
+    @Suppress("ForbiddenComment")
+    private fun ChildEntry<Routing>.suspend(): ChildEntry.Suspended<Routing> =
+        when (this) {
+            is ChildEntry.Suspended -> this
+            is ChildEntry.Initialized ->
+                ChildEntry.Suspended(
+                    key = key,
+                    // TODO: Not able to get a scope from Compose here, providing fake one
```
LachlanMcKee, line 205: ticket? What's the impact?

core/src/main/kotlin/com/bumble/appyx/core/children/ChildNodeCreationManager.kt
```diff
@@ -15,88 +16,141 @@ import kotlinx.coroutines.flow.update
...
-                                childMode = childMode,
-                            )
+            parentNode.navModel.screenState.collect { state ->
+                val navModelOnScreenKeys: Set<RoutingKey<Routing>>
+                val navModelOffScreenKeys: Set<RoutingKey<Routing>>
+                val navModelKeys: Set<RoutingKey<Routing>>
+                when (keepMode) {
+                    ChildEntry.KeepMode.KEEP -> {
+                        // Consider everything as on-screen for keep mode
+                        navModelOnScreenKeys =
```
LachlanMcKee, line 52: Perhaps this variable name is incorrect if we treat offScreen as onScreen? How about `navModelAvailableKeys` or `navModelKeptKeys`? (kept being the past tense of keep)

LachlanMcKee: If we change this, we should also change `updateChildren`

core/src/main/kotlin/com/bumble/appyx/core/children/ChildNodeCreationManager.kt
```diff
@@ -15,88 +16,141 @@ import kotlinx.coroutines.flow.update
...
+                            (state.onScreen + state.offScreen).mapNotNullToSet { element -> element.key }
+                        navModelOffScreenKeys = emptySet()
```
LachlanMcKee, line 54: `navModelOffScreenKeys` -> `navModelUnavailableKeys`

core/src/main/kotlin/com/bumble/appyx/core/children/ChildNodeCreationManager.kt
```diff
@@ -105,8 +159,59 @@ internal class ChildNodeCreationManager<Routing : Any>(
...
+                    key = key,
+                    // TODO: Not able to get a scope from Compose here, providing fake one
+                    savedState = node.saveInstanceState { true },
+                )
+        }
+
     private companion object {
         const val KEY_CHILDREN_STATE = "ChildrenState"
+
+        private fun <T, R : Any> Collection<T>.mapNotNullToSet(mapper: (T) -> R?): Set<R> =
```
LachlanMcKee, line 213: inline?

CherryPerry: Why? JVM can inline it by itself if required.

LachlanMcKee: That assumes a user of this library is using R8. I don't mind if you don't inline. though a lot of these types of functions in the kotlin lib are inlined

core/src/main/kotlin/com/bumble/appyx/core/children/ChildEntry.kt
CherryPerry: It a question and an answer is a value of enum.

core/src/main/kotlin/com/bumble/appyx/debug/Appyx.kt
CherryPerry: Do not have any plans, impact is miserable.

core/src/main/kotlin/com/bumble/appyx/core/children/ChildNodeCreationManager.kt
CherryPerry: Depends on a client usage of `SaverScope`. We are providing `SaverScope` right now do be able to use `canSave` in code, but honestly I do not see any value in it right now because we are supporting Android only. Also not sure if anybody ever used `canSave` except Compose internals.

LachlanMcKee (approved):

PR #125, opened 2022-09-02
Added parameter to generate jetpack compose compiler report

KovalevAndrey:
## Description

Added ability to generate report by compose compiler. 

Run `./gradlew assembleRelease -PenableComposeCompilerReports=true`

Report can be found in `../build/compose-metrics` of associated module

## Check list


build.gradle.kts
```diff
@@ -27,3 +29,23 @@ allprojects {
...
 }
+
+subprojects {
+    tasks.withType<KotlinCompile>().all {
+        kotlinOptions {
+            if (project.findProperty("enableComposeCompilerReports") == "true") {
+                freeCompilerArgs += listOf(
+                    "-P",
+                    "plugin:androidx.compose.compiler.plugins.kotlin:reportsDestination=" +
+                            project.buildDir.absolutePath + "/compose_metrics"
```
CherryPerry, line 40: Not safe path builder. File separators depend on OS.
Probably does not handles file path with spaces.

```kotlin
freeCompilerArgs += listOf(
  "-P",
  "plugin:androidx.compose.compiler.plugins.kotlin:reportsDestination=\"${File(project.buildDir, "compose_metrics")}\""
)
```

CherryPerry (changes requested):

build.gradle.kts
CherryPerry: Not sure about `absolutePath`, probably relational path to the project folder will be enough, but I am fine with absolute too.

CherryPerry (approved):

Issue #101, opened 2022-08-25
MviCoreExampleNode should not accept children connectors


CherryPerry:
`MviCoreExampleNode` should not provide connectors for children, but use some other mechanism to make children produce a result like we have in RIBs.

PR #128, opened 2022-09-05
Fix node connector invalid usage and add paren extensions

KovalevAndrey:
## Description

Fixes https://github.com/bumble-tech/appyx/issues/101

Also fixes cash in MviCoreExampleNode. The same instance of node connection was used for different instances of Nodes. This incorrect usage of `NodeConnectors` and led to a crash with error `Already flushed`

## Check list

### DIFF
--- a/core/src/main/kotlin/com/bumble/appyx/core/node/ParentNodeExt.kt
+++ b/core/src/main/kotlin/com/bumble/appyx/core/node/ParentNodeExt.kt
@@ -1,0 +1,15 @@
+package com.bumble.appyx.core.node
+
+import com.bumble.appyx.core.children.nodeOrNull
+
+fun ParentNode<*>.children(): List<Node> {
+    return this.children.value.values.mapNotNull { it.nodeOrNull }
+}
+
+inline fun <reified N : Node> ParentNode<*>.childrenOfType(): List<N> {
+    return this.children.value.values.mapNotNull { it.nodeOrNull }.filterIsInstance<N>()
+}
+
+inline fun <reified N : Node> ParentNode<*>.firstChildOfType(): N {
+    return childrenOfType<N>().first()
+}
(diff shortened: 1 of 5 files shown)


core/src/main/kotlin/com/bumble/appyx/core/node/ParentNodeExt.kt
```diff
@@ -0,0 +1,15 @@
+package com.bumble.appyx.core.node
+
+import com.bumble.appyx.core.children.nodeOrNull
+
+fun ParentNode<*>.children(): List<Node> {
+    return this.children.value.values.mapNotNull { it.nodeOrNull }
+}
+
+inline fun <reified N : Node> ParentNode<*>.childrenOfType(): List<N> {
+    return this.children.value.values.mapNotNull { it.nodeOrNull }.filterIsInstance<N>()
```
CherryPerry, line 10: With 1 mapping – `values.mapNotNull { it.nodeOrNull as? T }`

core/src/main/kotlin/com/bumble/appyx/core/node/ParentNodeExt.kt
```diff
@@ -0,0 +1,15 @@
...
+}
+
+inline fun <reified N : Node> ParentNode<*>.firstChildOfType(): N {
+    return childrenOfType<N>().first()
```
CherryPerry, line 14: Here we can avoid list allocation by using `firstOrNull { it.nodeOrNull as? N }`

Not sure, should we crash or not in case of missing child? Maybe make nullable?

CherryPerry (approved):

LachlanMcKee:
@KovalevAndrey it looks like you added some extension functions to the library. It would be good to mention this in the change log

Issue #129, opened 2022-09-07
Remove sealed interface from operations and allow clients to define their own


KovalevAndrey:
Some operations are defined as sealed for having exhaustive when conditions when implementing transition handler. We should allow customers to define their own operations for the existing NavModels. 

Discussion for the reference: https://github.com/bumble-tech/appyx/discussions/123

PR #130, opened 2022-09-07
Remove sealed interface from operations

KovalevAndrey:
## Description

Fixes https://github.com/bumble-tech/appyx/issues/129

## Check list


core/src/main/kotlin/com/bumble/appyx/navmodel/backstack/transitionhandler/BackStackFader.kt
```diff
@@ -8,16 +9,17 @@ import androidx.compose.runtime.remember
...
 import com.bumble.appyx.core.navigation.transition.TransitionDescriptor
 import com.bumble.appyx.core.navigation.transition.TransitionSpec
+import com.bumble.appyx.navmodel.backstack.BackStack
 
 @Suppress("TransitionPropertiesLabel")
 class BackStackFader<T>(
     private val transitionSpec: TransitionSpec<BackStack.TransitionState, Float> = { spring() }
 ) : ModifierTransitionHandler<T, BackStack.TransitionState>() {
 
+    @SuppressLint("ModifierFactoryExtensionFunction")
```
CherryPerry, line 22: Maybe we should change API instead of suppressing it?
Also makes sense to regenerate lint.xml

KovalevAndrey: yes, but not in the scope of this issue. I'll update lint-baseline

CherryPerry (approved):

PR #138, opened 2022-09-09
Introduced dependency-analysis and removed unused deps

LachlanMcKee:
## Description
Introduces dependency-analysis which allows us to detect ABI changes as well as unused dependencies.
https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin

## Check list


core/build.gradle.kts
```diff
@@ -61,5 +59,6 @@ dependencies {
     androidTestImplementation(libs.androidx.test.espresso.core)
     androidTestImplementation(libs.androidx.test.junit)
     androidTestImplementation(libs.compose.ui.test.junit4)
+    androidTestImplementation(libs.compose.foundation)
```
LachlanMcKee, line 62: We only use foundation APIs (in this case Modifier.background) in our instrumentation tests

sandbox/build.gradle.kts
```diff
@@ -61,19 +61,17 @@ dependencies {
...
     implementation(libs.compose.material)
     implementation(libs.compose.ui.tooling)
     implementation(libs.compose.ui.ui)
     implementation(libs.google.accompanist.flow)
     implementation(libs.google.material)
-    implementation(libs.mvicore)
-    implementation(libs.ribs.base)
-    implementation(libs.ribs.compose)
-    implementation(libs.ribs.rx)
+    implementation(libs.ribs.mvicore)
```
LachlanMcKee, line 71: `ribs.mvicore` exposes `ribs.base`, `ribs.compose` and `ribs.rx` using `api`

testing-junit4/build.gradle.kts
```diff
@@ -33,7 +33,6 @@ android {
 
 dependencies {
     api(project(":testing-unit-common"))
-    api(libs.androidx.test.rules)
```
LachlanMcKee, line 36: This never made sense in the context of a non-Android testing module.

testing-junit5/build.gradle.kts
```diff
@@ -38,6 +38,5 @@ android {
 dependencies {
     api(project(":testing-unit-common"))
     implementation(libs.junit.api)
-    implementation(libs.androidx.arch.core.testing)
```
LachlanMcKee, line 41: This never made sense in the context of a non-Android testing module.

testing-unit-common/build.gradle.kts
```diff
@@ -33,8 +33,5 @@ android {
 
 dependencies {
     api(project(":core"))
-
-    implementation(libs.androidx.lifecycle.runtime)
```
LachlanMcKee, line 37: This is exposed via core as it uses `api(libs.androidx.appcompat)`

PR #141, opened 2022-09-10
[#140] Created testing-ui-activity to avoid leaking test code in prod

### LINKED ISSUE #140: testing-ui module code is currently being included in sandbox release build
The `testing-ui` module is currently being defined as an `implementation` as well as an `androidTestImplementation` dependency. When I attempted to remove the `implementation` usage, the tests started failing.

This is because the `testing-ui` module includes an activity. This activity must be registered within the main apps manifest. This would then require developers to do something like `debugImplementation testing-ui` which then leaks test code into the debug build.

To fix this, we should create a `testing-ui-activity` module which only defines the activity. This allows the tests to keep working, and also avoids adding espresso code in a debug build.

This would like like:
`debugImplementation testing-ui-activity`
`androidTestImplementation testing-ui`

See Google's own documentation that shows that this is necessary: https://developer.android.com/jetpack/compose/testing#setup

I tried to get the activity working without adding it into the main apk, and it doesn't seem to work.

LachlanMcKee:
## Description
Fixes #140

## Check list


CherryPerry:
BTW no need to create an issue just to open a PR straight away, you can describe the issue here directly.

CherryPerry (approved):
Can you also investigate how it works without this change in our apps please?

LachlanMcKee:
@CherryPerry will do!

LachlanMcKee:
@CherryPerry from what I can see there should be no issues based on how our existing tests are structured. however I will keep this in mind when we next update

PR #143, opened 2022-09-11
Fix gradle transitive deps

LachlanMcKee:
## Description
(Depends on #138)

Correctly exposes transitive dependencies that are part of the libraries ABI. Previously if someone used one of these libraries in isolation, they could have faced quite a few problematic compilation issues.

## Check list

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,9 @@
 
 ## Pending changes
 
-- [#138](https://github.com/bumble-tech/appyx/pull/139) - **Fixed**: `androidx.appcompat:appcompat` from is exposed via `api` within `com.bumble.appyx:core`. This prevents potential compilation bugs.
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 ## Pending changes
 
 - [#138](https://github.com/bumble-tech/appyx/pull/139) - **Fixed**: `androidx.appcompat:appcompat` from is exposed via `api` within `com.bumble.appyx:core`. This prevents potential compilation bugs.
```
CherryPerry, line 6: Both changes has wrong links (numbers).

CherryPerry (approved):

CHANGELOG.md
LachlanMcKee: good catch. will fix.

LachlanMcKee: fixed! also added the link to @KovalevAndrey's PR

Issue #146, opened 2022-09-12
Remove FragmentIntegrationPoint 


KovalevAndrey:
FragmentIntegrationPoint is no longer needed as we should use `ActivityIntegrationPoint.getIntegrationPoint(context: Context)` method.

PR #147, opened 2022-09-12
Remove fragment integration point

KovalevAndrey:
## Description

Fixes https://github.com/bumble-tech/appyx/issues/146

## Check list


zsoltk (approved):

LachlanMcKee:
Are you sure that this will work as expected? There is some logic in this integration point about navigating up etc that. Perhaps removing this is good as it leaves that decision up to users?

KovalevAndrey:
> Are you sure that this will work as expected? There is some logic in this integration point about navigating up etc that. Perhaps removing this is good as it leaves that decision up to users?

FragmentIntegrationPoint effectively does the same as ActivityIntegrationPoint but call calling `val activity = fragment.requireActivity()` before action so I don't change the behaviour here

CherryPerry (approved):

LachlanMcKee (approved):
Thanks for the clarification

PR #154, opened 2022-09-18
WIP - Updated androidx and accompanist libraries

LachlanMcKee:
## Description
Testing same change as the other PR, but from the main repo. This will help us see where the permissions issue exists for dependency diff

## Check list


LachlanMcKee:
It looks like the diff is too large. I might need to tweak it so that it only shows the first couple of dependencies, and also truncates it if it gets too large

Issue #142, opened 2022-09-10
Make ActivityIntegrationPoint use Lifecycle


CherryPerry:
Make `ActivityIntegrationPoint` use `Lifecycle` instead of weak references.

```kotlin
activity.lifecycle.addObserver(
  onDestroy = { integrationPointsMap.remove(activity) }
)
```

Related to https://github.com/bumble-tech/appyx/pull/139 https://github.com/bumble-tech/appyx/pull/104

PR #157, opened 2022-09-20
Remove WeakReference from IntegrationPoint and explicitly remove activity from

KovalevAndrey:
## Description

Fixes https://github.com/bumble-tech/appyx/issues/142

## Check list

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 
 ## Pending changes
 
+- [#142](https://github.com/bumble-tech/appyx/issues/142) - **Breaking change**: ActivityIntegrationPoint uses ComponentActivity and explicitly removes entity from Map in onDestroy
 - [#146](https://github.com/bumble-tech/appyx/issues/146) - **Breaking change**: Removed `FragmentIntegrationPoint`. Clients should use `ActivityIntegrationPoint.getIntegrationPoint(context: Context)` to get integration point from Fragment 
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/com/bumble/appyx/core/integrationpoint/ActivityIntegrationPoint.kt
```diff
@@ -50,14 +51,22 @@ open class ActivityIntegrationPoint private constructor(
...
             savedInstanceState: Bundle?,
         ): ActivityIntegrationPoint =
             ActivityIntegrationPoint(activity, savedInstanceState)
-                .also { integrationPoints[activity] = WeakReference(it) }
+                .also {
+                    integrationPoints[activity] = it
+
+                    activity.lifecycle.addObserver(object : DefaultLifecycleObserver {
+                        override fun onDestroy(owner: LifecycleOwner) {
+                            integrationPoints.remove(activity)
```
LachlanMcKee, line 66: should we remove the observer here?

LachlanMcKee: We should run some memory leak analysis on this change to be sure that we are not introducing a memory leak.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 
 ## Pending changes
 
+- [#142](https://github.com/bumble-tech/appyx/issues/142) - **Breaking change**: ActivityIntegrationPoint uses ComponentActivity and explicitly removes entity from Map in onDestroy
```
LachlanMcKee, line 6: We definitely need to add this to our documentation somewhere.

i.e. `We do not support accessing the integration point within onDestroy`

PR #162, opened 2022-09-21
Added disposeOnDestroyPlugin to interop-rx2

LachlanMcKee:
## Description
Added disposeOnDestroyPlugin to interop-rx2. This will allow Rx2 code to be easily disposed when the node it belongs to is destroyed.

## Check list

### DIFF
--- a/interop-rx2/src/main/kotlin/com/bumble/appyx/interop/rx2/plugin/DisposeOnDestroyPlugin.kt
+++ b/interop-rx2/src/main/kotlin/com/bumble/appyx/interop/rx2/plugin/DisposeOnDestroyPlugin.kt
@@ -1,0 +1,17 @@
+package com.bumble.appyx.interop.rx2.plugin
+
+import com.bumble.appyx.core.plugin.Destroyable
+import com.bumble.appyx.core.plugin.Plugin
+import io.reactivex.disposables.CompositeDisposable
+import io.reactivex.disposables.Disposable
+
+private class DisposeOnDestroyPlugin(disposables: List<Disposable>) : Destroyable {
+    private val disposable = CompositeDisposable(disposables)
+
+    override fun destroy() {
+        disposable.dispose()
+    }
+}
+
+fun disposeOnDestroyPlugin(vararg disposables: Disposable): Plugin =
+    DisposeOnDestroyPlugin(disposables.toList())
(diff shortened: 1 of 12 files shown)


interop-rx2/src/main/kotlin/com/bumble/appyx/interop/rx2/plugin/DisposeOnDestroyPlugin.kt
```diff
@@ -0,0 +1,17 @@
...
+
+private class DisposeOnDestroyPlugin(disposables: List<Disposable>) : Destroyable {
+    private val disposable = CompositeDisposable(disposables)
+
+    override fun destroy() {
+        disposable.dispose()
+    }
+}
+
+fun disposeOnDestroyPlugin(vararg disposables: Disposable): Plugin =
```
LachlanMcKee, line 16: I deliberately didn't call it `disposables` like in RIBs because I feel that the extension function needs more context - especially since it isn't scoped.

CherryPerry (approved):

LachlanMcKee:
FYI @CherryPerry I made an extra change to the PR after realising that we needed to use this function within our sandbox code (to show best practices in case someone uses it as a reference)

testing-unit-common/src/main/kotlin/com/bumble/appyx/testing/unit/common/helper/NodeTestHelper.kt
```diff
@@ -19,11 +19,6 @@ open class NodeTestHelper<N : Node>(private val node: N) {
         require(state != Lifecycle.State.INITIALIZED) {
             "Can't move to INITIALIZED state"
         }
-        require(
```
LachlanMcKee, line 22: Being able to move to DESTROYED is a valid test case, and the Node class itself has a safeguard to prevent moving from a destroyed state to another state.

interop-rx2/src/main/kotlin/com/bumble/appyx/interop/rx2/plugin/DisposeOnDestroyPlugin.kt
```diff
@@ -0,0 +1,17 @@
+package com.bumble.appyx.interop.rx2.plugin
+
+import com.bumble.appyx.core.plugin.Destroyable
+import com.bumble.appyx.core.plugin.Plugin
+import io.reactivex.disposables.CompositeDisposable
+import io.reactivex.disposables.Disposable
+
+private class DisposeOnDestroyPlugin(disposables: List<Disposable>) : Destroyable {
```
KovalevAndrey, line 8: we don't use Plugin suffix for other plugin, let's make it consistent?

KovalevAndrey (approved):

interop-rx2/src/main/kotlin/com/bumble/appyx/interop/rx2/plugin/DisposeOnDestroyPlugin.kt
LachlanMcKee: We probably should. Though as this is private, I'm happy to do this for :)

PR #169, opened 2022-09-26
Updated accompanist flowlayout to match Compose version

LachlanMcKee:
## Description
Updated accompanist flowlayout to a version that uses the same Compose version.

According to their documentation, there are no functional changes to the library other than possibly minor changes to make it work with Compose 1.2.1.

## Check list


gradle/libs.versions.toml
```diff
@@ -40,7 +40,7 @@ compose-runtime = { module = "androidx.compose.runtime:runtime", version.ref = "
 compose-ui-test-junit4 = { module = "androidx.compose.ui:ui-test-junit4", version.ref = "compose" }
 compose-ui-tooling = { module = "androidx.compose.ui:ui-tooling", version.ref = "compose" }
 compose-ui-ui = { module = "androidx.compose.ui:ui", version.ref = "compose" }
-google-accompanist-flow = "com.google.accompanist:accompanist-flowlayout:0.20.0"
+google-accompanist-flow = "com.google.accompanist:accompanist-flowlayout:0.25.1"
```
LachlanMcKee, line 43: This is only used in our sandbox, so I don't think it's worth calling out in the changelog

CherryPerry (approved):

Issue #98, opened 2022-08-24
Drop com.bumble.appyx.core.minimal.reactive


CherryPerry:
We already have `attachWorkflow` and `waitForChildAttached` coroutine powered public API, so I think we can migrate copy-pasted from RIBs `PermissionRequestBoundary` to Flows.

Or we should commit to `minimal.reactive` and change `attachWorkflow`, `waitForChildAttached` implementations.

@zsoltk @KovalevAndrey @LachlanMcKee

LachlanMcKee:
I agree that it would make sense to remove this, otherwise we should be adding reactive support for everything to be consistent. I don't think that is a good use of our time at the moment.

PR #173, opened 2022-09-27
Remove custom rx

CherryPerry:
## Description

Fixes https://github.com/bumble-tech/appyx/issues/98

Existing code if it uses `events(code).rx2()` should not break.
Clients should manually add `kotlinx-coroutines-rx2` as a dependency if they need it.

## Check list


zsoltk (approved):

sandbox/src/main/kotlin/com/bumble/appyx/sandbox/client/integrationpoint/IntegrationPointExampleNode.kt
```diff
@@ -129,10 +125,10 @@ class IntegrationPointExampleNode(buildContext: BuildContext) : Node(buildContex
...
     }
 
-    private fun observeActivityResult() {
-        activityResultsCancellable = activityStarter
-            .events(this)
-            .observe { event ->
+    private suspend fun observeActivityResult() {
+        activityStarter
+            .events(this@IntegrationPointExampleNode)
+            .collect { event ->
```
KovalevAndrey, line 131: can you please fix `requestCodeClientId`. It will crash after requesting permissions + rotate screen.

`override val requestCodeClientId: String = IntegrationPointExampleNode::class.java.name`

KovalevAndrey (approved):

Issue #174, opened 2022-09-27
IntegrationPointExample does not work with "do not keep activities"


CherryPerry:
Steps:
1. Do not keep activities
2. Open IntegrationPointExample
3. Launch activity for result
4. Return random result code
5. Node has not received it

Issue:
1.  The node is created and subscribed to `ActivityStarter`
2. When the next activity is launched the node is destroyed
3. `onActivityResult` is invoked and events are pushed
4. Compose rendering is triggered and the node is recreated (`Activity.setContent {  }`)
5. The node is created and subscribed to `ActivityStarter`

Discovered after converting error log into exception in https://github.com/bumble-tech/appyx/pull/173

PR #178, opened 2022-09-29
Cache integration point result if there are no subscribers and flush when they appear

KovalevAndrey:
## Description

Fixes https://github.com/bumble-tech/appyx/issues/174

## Check list

### DIFF
--- a/libraries/core/src/main/kotlin/com/bumble/appyx/core/integrationpoint/requestcode/RequestCodeBasedEventStreamImpl.kt
+++ b/libraries/core/src/main/kotlin/com/bumble/appyx/core/integrationpoint/requestcode/RequestCodeBasedEventStreamImpl.kt
@@ -1,14 +1,21 @@
 package com.bumble.appyx.core.integrationpoint.requestcode
 
-import com.bumble.appyx.Appyx
 import com.bumble.appyx.core.integrationpoint.requestcode.RequestCodeBasedEventStream.RequestCodeBasedEvent
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Job
 import kotlinx.coroutines.channels.BufferOverflow
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.collectLatest
+import kotlinx.coroutines.launch
+import kotlin.coroutines.EmptyCoroutineContext
 
 abstract class RequestCodeBasedEventStreamImpl<T : RequestCodeBasedEvent>(
-    private val requestCodeRegistry: RequestCodeRegistry
+    private val requestCodeRegistry: RequestCodeRegistry,
+    protected val scope: CoroutineScope = CoroutineScope(EmptyCoroutineContext + Dispatchers.Unconfined)
 ) : RequestCodeBasedEventStream<T> {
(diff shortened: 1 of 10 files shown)


libraries/core/src/main/kotlin/com/bumble/appyx/core/integrationpoint/requestcode/RequestCodeBasedEventStreamImpl.kt
```diff
@@ -35,23 +42,39 @@ abstract class RequestCodeBasedEventStreamImpl<T : RequestCodeBasedEvent>(
         }
     }
 
+    private fun cacheResultAndFlushOnNewSubscribers(id: Int, flow: MutableSharedFlow<T>, event: T) {
+        cache[id] = event
+        var job: Job? = null
+        job = scope.launch {
+            flow.subscriptionCount.collectLatest { subscriptionCount ->
+                if (subscriptionCount > 0) {
+                    val cachedValue = cache[id]
```
CherryPerry, line 51: Do you need `cache`? Looks like you can capture it in lambda.

CherryPerry (approved):

libraries/core/src/main/kotlin/com/bumble/appyx/core/integrationpoint/requestcode/RequestCodeBasedEventStreamImpl.kt
```diff
@@ -35,23 +41,33 @@ abstract class RequestCodeBasedEventStreamImpl<T : RequestCodeBasedEvent>(
         }
     }
 
+    private fun flushResultWhenSubscribersAppear(flow: MutableSharedFlow<T>, event: T) {
```
LachlanMcKee, line 44: I think in a future PR we should investigate whether we need to remove flows from the events map in cases where a node is never recreated

LachlanMcKee (approved):

libraries/core/src/main/kotlin/com/bumble/appyx/core/node/Node.kt
```diff
@@ -74,9 +76,12 @@ abstract class Node(
         }
 
     private val lifecycleRegistry = LifecycleRegistry(this)
-
     private var wasBuilt = false
 
+    override val requestCodeClientId: String =
+        buildContext.savedStateMap?.get(RequestCodeClientIdKey) as String?
```
CherryPerry, line 82: Let's introduce public `id` property for node.

libraries/core/src/main/kotlin/com/bumble/appyx/core/node/Node.kt
```diff
@@ -211,6 +216,7 @@ abstract class Node(
         plugins<UpNavigationHandler>().any { it.handleUpNavigation() }
 
     companion object {
+        private const val RequestCodeClientIdKey = "node.requestCodeClientId"
```
CherryPerry, line 219: FULL_CAPS as in other places? It is not Compose code.

libraries/core/src/main/kotlin/com/bumble/appyx/core/node/Node.kt
CherryPerry: https://github.com/bumble-tech/appyx/pull/175/files#diff-a57bab82c13427fb82ff9ac8efdc79f337a988817310bb19a006a9a276df04ddR54

CherryPerry (approved):

PR #180, opened 2022-09-29
Ensure onSavedInsatnceState super called

KovalevAndrey:
## Description

Ensure onSavedInsatnceState was called for restoring Node's state correctly 

## Check list


CherryPerry:
Not sure if we need it. Have you experienced issues with this behaviour?

KovalevAndrey:
> Not sure if we need it. Have you experienced issues with this behaviour?

I think we need to make sure that integration point works correctly. If clients override it and don't call super - it won't

LachlanMcKee (approved):

CherryPerry (approved):

PR #184, opened 2022-10-03
Fix Detekt issues and modify rules

mapm14:
## Description

Fix Detekt issues and modify rules

## Check list

### DIFF
--- a/detekt.yml
+++ b/detekt.yml
@@ -2,16 +2,12 @@ config:
   warningsAsErrors: true
 
 complexity:
-  ComplexMethod:
-    threshold: 17
   LongMethod:
     threshold: 70
   LongParameterList:
     # It is suggested to increase this for Compose: https://detekt.dev/docs/introduction/compose/#longparameterlist
     functionThreshold: 8
     ignoreDefaultParameters: true
-  TooManyFunctions:
-    thresholdInClasses: 18
 
 naming:
   active: true
@@ -23,8 +19,7 @@ style:
   ForbiddenComment:
     values: [ 'FIXME:', 'STOPSHIP:' ]
   MagicNumber:
-    ignoreNumbers: [ '-1', '0', '1', '2', '3' ]
-    excludes: [ '**/*TransitionHandler.kt', '**/*Test.kt' ]
+    excludes: [ '**/*Test.kt' ]
   MaxLineLength:
     excludes: [ '**/test/**', '**/*.Test.kt', '**/*.Spec.kt' ]
     excludeCommentStatements: true
--- a/detekt.yml
+++ b/detekt.yml
@@ -17,7 +17,7 @@ naming:
 
 style:
   ForbiddenComment:
-    values: [ 'FIXME:', 'STOPSHIP:' ]
+    allowedPatterns: 'https://github.com/bumble-tech/appyx/issues/*'
   MagicNumber:
     excludes: [ '**/*Test.kt' ]
   MaxLineLength:
--- a/detekt.yml
+++ b/detekt.yml
@@ -2,8 +2,6 @@ config:
   warningsAsErrors: true
 
 complexity:
-  LongMethod:
-    threshold: 70
   LongParameterList:
     # It is suggested to increase this for Compose: https://detekt.dev/docs/introduction/compose/#longparameterlist
     functionThreshold: 8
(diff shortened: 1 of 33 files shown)


detekt.yml
```diff
@@ -2,24 +2,35 @@ config:
   warningsAsErrors: true
 
 complexity:
+  ComplexMethod:
```
LachlanMcKee, line 5: I would prefer that we don't change these values just so our existing code passes. We should leave the thresholds as they were to encourage us to not create large, complicated and potentially unreadable code.

detekt.yml
```diff
@@ -2,24 +2,35 @@ config:
...
+    threshold: 17
+  LongMethod:
+    threshold: 70
   LongParameterList:
     # It is suggested to increase this for Compose: https://detekt.dev/docs/introduction/compose/#longparameterlist
     functionThreshold: 8
     ignoreDefaultParameters: true
+  TooManyFunctions:
+    thresholdInClasses: 18
```
LachlanMcKee, line 14: Same comment as above

detekt.yml
```diff
@@ -2,24 +2,35 @@ config:
...
 naming:
   active: true
   FunctionNaming:
     active: true
-    ignoreAnnotated: ['Composable']
+    ignoreAnnotated: [ 'Composable' ]
 
 style:
+  ForbiddenComment:
+    values: [ 'FIXME:', 'STOPSHIP:' ]
```
LachlanMcKee, line 24: What's wrong with leaving TODO as a forbidden comment?

mapm14: I put it back and added `https://github.com/bumble-tech/appyx/issues/*` as an allowed pattern

detekt.yml
```diff
@@ -2,6 +2,8 @@ config:
   warningsAsErrors: true
 
 complexity:
+  LongMethod:
+    threshold: 70
```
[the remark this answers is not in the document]
LachlanMcKee: Can you not just suppress them? The discussion we had yesterday was that yes, it's okay in Compose if we go over this limit, but non-compose code should not.

By setting this threshold to 70, we will potentially allow functions in non-compose code that are too complicated.

libraries/core/src/test/kotlin/com/bumble/appyx/core/children/ChildAwareCommonTestSpec.kt
```diff
@@ -150,6 +150,7 @@ interface ChildAwareCommonTestSpec {
         assertEquals(setOf(children[0], children[1]), capturedNodes2)
     }
 
+    @Suppress("MagicNumber")
```
LachlanMcKee, line 153: Is this needed? It looks like you suppressed all magic numbers in tests

samples/app/src/main/kotlin/com/bumble/appyx/app/node/onboarding/OnboardingContainerNode.kt
```diff
@@ -169,12 +154,36 @@ class OnboardingContainerNode(
...
                         }
-                        Spacer(Modifier)
                     }
                 }
             }
         }
     }
+
+    @Composable
+    private fun PreviousAndNextButtons(
```
LachlanMcKee, line 164: I believe this change might go against the Compose Lint rules that we will likely introduce in the future. The problem is that there are now multiple composables being created without a specified scope (Column/Row), so calling this could have unindented behaviour

LachlanMcKee: I guess we can resolve that once we introduce the lint rules.

samples/navmodel-samples/detekt.yml
```diff
@@ -0,0 +1,4 @@
+style:
+  MagicNumber:
+    # Magic numbers are okay for sample apps.
+    active: false
```
LachlanMcKee, line 4: nice!

samples/sandbox/src/main/kotlin/com/bumble/appyx/sandbox/client/backstack/BackStackExampleNode.kt
```diff
@@ -140,183 +144,250 @@ class BackStackExampleNode(
...
+                    backStackState = backStackState,
+                )
+                BackstackColumn(backStackState = backStackState)
+                ColumnSkipRendering(skipChildRenderingByNavTarget = skipChildRenderingByNavTarget)
+            }
+        }
+    }
+
+    @Composable
+    private fun ChildColumn(
```
LachlanMcKee, line 181: Similar comment here about not having a scope (which could cause issues with the eventual Compose lint rules)

samples/app/src/main/kotlin/com/bumble/appyx/app/node/onboarding/OnboardingContainerNode.kt
mapm14: I'm not familiar with that Compose Lint rule, you mean just to do `private fun RowScope.PreviousAndNextButtons`?
If that's the case I can do it now

LachlanMcKee (approved):

PR #186, opened 2022-10-04
Portal lifecycle problem ideation

KovalevAndrey:
## Description

1. When portaled node enters composition report to parent
2. Parent caps all children lifecycles except for the portal node
3. Repeat step 2 until we reach the root of the tree 

## Check list


libraries/core/src/main/kotlin/com/bumble/appyx/core/node/ParentNode.kt
```diff
@@ -89,6 +91,19 @@ abstract class ParentNode<NavTarget : Any>(
         manageTransitions()
     }
 
+    fun onPortalRemoved(node: Node) {
+        parent?.onPortalRemoved(this)
+        childNodeLifecycleManager.removePortal()
```
zsoltk, line 96: Double check if the assumption that we can remove the last one w/o a reference will always be correct

PR #188, opened 2022-10-04
Immediately complete transitions pending on a Child on removal from composition

Ninjars:
Courtesy of a suggestion by Yury.

## Description

Implemented DisposableEffect in Child to complete transitions when it leaves composition rather than leaving them hanging. This prevents unexpected transition animations from occurring when, for example, scrolling a LazyList containing elements that may have been mid-transition when the user scrolled past them, and avoids situations where subsequent transitions may get into races to complete with pending ones.

## Check list

### DIFF
--- a/libraries/core/src/main/kotlin/com/bumble/appyx/Appyx.kt
+++ b/libraries/core/src/main/kotlin/com/bumble/appyx/Appyx.kt
@@ -6,6 +6,8 @@ object Appyx {
 
     var exceptionHandler: ((Exception) -> Unit)? = null
     var defaultChildKeepMode: ChildEntry.KeepMode = ChildEntry.KeepMode.KEEP
+    var defaultChildTransitionStrategy: ChildEntry.ChildTransitionStrategy =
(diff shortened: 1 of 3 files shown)


CherryPerry:
Let's postpone merge and release after 1.0. Maybe in 2.0. Not sure about side effects of this change, need some time to find out.

LachlanMcKee:
@CherryPerry perhaps we could include this in 1.x but have it as an optional behaviour? (false by default?) At least that would allow developers to use it and provide feedback about whether it's a good idea or not.

Ninjars:
maybe enabled by default as a behaviour change that can be disabled if it's discovered to cause an issue? The better to get feedback sooner?

KovalevAndrey (changes requested):
This contradicts with the way we handle draw children in `Children` composable. We do it the other way around: first we call `onTransitionFinished`, then `NavModel` state is updated, then composable is disposed. I'm not sure this logic should be applied to every `Child` function. I'd say it should be handled on the `DrawChildrenInTheList` composable function which draws children not a single container, but it the list

PR #189, opened 2022-10-04
Branch split

CherryPerry:
## Description

CI setup for split.

Using `./gradlew` instead of the action now to not copy-paste cache disabling config everywhere. Should work, this use case is described in the action docs.

## Check list


.github/workflows/build.yml
```diff
@@ -20,33 +24,35 @@ jobs:
...
+          --continue
+      - name: Check publication setup
+        run: >
+          ./gradlew
+          publishAppyxReleasePublicationToOSSRHRepository
+          publishAppyxReleasePublicationToSonatypeSnapshotRepository
+          --dry-run
+          --no-parallel
+      - name: Deploy snapshot
+        if: env.MAIN_BRANCH == 'true' && github.repository == 'bumble-tech/appyx'
```
LachlanMcKee, line 45: This could be problematic, the 1.0 and 2.0 snapshots will override each other. Perhaps we need to also change the snapshot version to handle 1.x and 2.x branches?

.github/workflows/build.yml
```diff
@@ -67,18 +73,19 @@ jobs:
   instrumentation-tests:
     name: Instrumentation tests
     runs-on: macOS-latest
-    timeout-minutes: 60
+    timeout-minutes: 30
```
LachlanMcKee, line 76: Are we definitely always below 30 minutes?

.github/workflows/build.yml
```diff
@@ -20,33 +24,35 @@ jobs:
...
-      - uses: gradle/gradle-build-action@v2
-        with:
-          arguments: |
-            publishAppyxReleasePublicationToOSSRHRepository
-            publishAppyxReleasePublicationToSonatypeSnapshotRepository
-            --dry-run
-            --no-parallel
-      - uses: gradle/gradle-build-action@v2
-        if: github.ref == 'refs/heads/main' && github.repository == 'bumble-tech/appyx'
+          cache-read-only: ${{ env.MAIN_BRANCH != 'true' }}
```
LachlanMcKee, line 27: Are we okay with having two different `main` branches updating the cache? surely they will conflict with one another?

LachlanMcKee (dismissed):

.github/workflows/build.yml
CherryPerry: 15 mins in general

.github/workflows/build.yml
CherryPerry: I expect the cache to be merged.

.github/workflows/build.yml
CherryPerry: Will split.

.github/workflows/build.yml
LachlanMcKee: Is that ideal? what if the 2.x branch changes dependencies ahead of the 1.x branch? They could constantly override each other

CherryPerry: Tried to switch a version of Coil multiple times and to recompile sandbox, got cache hits for both Coil versions.

Maybe approach is not ideal, but I can't provide anything better, `gradle-build-action` does not allows to define additional keys.

LachlanMcKee: fair enough

LachlanMcKee (approved):

PR #197, opened 2022-10-08
Remove plugin implementation from view and retrieve node instance from LocalComposition

KovalevAndrey:
## Description

Remove plugin implementation from `View`. This will allow to merge `AppyxViewTestRule` and `AppyxParentViewTestRule`

## Check list


CherryPerry (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Pending changes
 
+- [#197](https://github.com/bumble-tech/appyx/pull/197) – **Breaking change**: `ParentNodeView` does not implement plugin anymore. Node instance is retrieved via LocalComposition.
```
LachlanMcKee: Maybe mention that `AbstractParentNodeView` and `AppyxParentViewTestRule.kt` have been removed?

libraries/core/src/main/kotlin/com/bumble/appyx/core/node/ParentNodeView.kt
```diff
@@ -2,9 +2,17 @@ package com.bumble.appyx.core.node
...
 
-interface ParentNodeView<NavTarget : Any> : NodeView, NodeAware<ParentNode<NavTarget>> {
+interface ParentNodeView<NavTarget : Any> : NodeView {
+
     @Composable
     fun ParentNode<NavTarget>.NodeView(modifier: Modifier)
+
+    @Suppress("UNCHECKED_CAST")
+    @Composable
+    override fun View(modifier: Modifier) {
```
LachlanMcKee, line 16: is this okay if it gets overriden? or should it be final?

KovalevAndrey: Ideally it should be final, the problem is that `ParentNodeVew` is an interface. (requested changes)

LachlanMcKee: Maybe add some JavaDoc specifying that overriding this function will cause unexpected behaviours?

LachlanMcKee (approved):

PR #200, opened 2022-10-09
Added and fixed twitter compose detekt rules

LachlanMcKee:
## Description
Added twitter compose detekt rules to ensure that new code that is contributed meets Composable guidelines.
https://twitter.github.io/compose-rules/

## Check list


libraries/core/src/main/kotlin/com/bumble/appyx/core/clienthelper/interactor/Interactor.kt
```diff
@@ -7,7 +7,7 @@ import com.bumble.appyx.core.plugin.NodeAware
 import com.bumble.appyx.core.plugin.NodeLifecycleAware
 import com.bumble.appyx.core.plugin.SavesInstanceState
 
-abstract class Interactor<N: Node>(
+open class Interactor<N: Node>(
```
LachlanMcKee, line 10: These classes don't have any abstract methods. Not sure why including the Compose detekt rules flagged these, but it doesn't hurt the API to change these.

documentation/faq.md
```diff
@@ -75,4 +75,4 @@ We're full with ideas where to take Appyx further! A more detailed roadmap will
 
 ## Other
 
-Have a question? Raise it in [Discussions](https://github.com/bumble-tech/appyx/discussions)!.
\ No newline at end of file
+Have a question? Raise it in [Discussions](https://github.com/bumble-tech/appyx/discussions)!.
```
LachlanMcKee, line 78: I accidentally reformatted these files, and because they didn't have a newline character at the end, Android Studio won't stop trying to stage them, so I thought I may as well commit the change

gradle/libs.versions.toml
```diff
@@ -17,6 +17,7 @@ rxjava2 = "2.2.21"
 rxandroid = "2.1.1"
 junit5 = "5.8.2"
 detekt = "1.21.0"
+detektCompose = "0.0.18"
```
CherryPerry, line 20: btw no need to extract version if it is not reused

CherryPerry (approved):

gradle/libs.versions.toml
LachlanMcKee: Good point

libraries/core/src/main/kotlin/com/bumble/appyx/core/node/ParentNode.kt
```diff
@@ -261,11 +261,13 @@ abstract class ParentNode<NavTarget : Any>(
 
     private class PermanentChildRender(private val node: Node) : ChildRenderer {
 
+        @Suppress("ComposableNaming")
         @Composable
         override fun invoke(modifier: Modifier) {
```
KovalevAndrey, line 266: Can you please add operator here?
```suggestion
        override operator fun invoke(modifier: Modifier) {
```

KovalevAndrey (approved):

libraries/core/src/main/kotlin/com/bumble/appyx/core/node/ParentNode.kt
LachlanMcKee: done

Issue #208, opened 2022-10-11
Make sample apps a single app


LachlanMcKee:
Currently our sample apps approach (creating a standalone app for each sample) is not scalable.

It is great to have each sample module as an isolated implementation of a particular approach (Jetpack compose navigation, dagger, etc). However what we should do to make this more scalable is to make each sample app a library module that then 'plugs' into the sample app.

This would mean we only have the user install one sample app, and we also don't need to have multiple different app icons.

PR #217, opened 2022-10-13
[#208] Moved existing samples into single sample app

LachlanMcKee:
## Description
Moved existing samples into single sample app.

I have recorded two videos demonstrating the behaviour. Note that I have disabled animations on my emulator (they are not broken)

**Sample app first use**
[samples-app-first-use.webm](https://user-images.githubusercontent.com/1333155/195810250-763df557-8d90-472c-9408-cd1869454e30.webm)

**Sample app second use**
[samples-app-second-use.webm](https://user-images.githubusercontent.com/1333155/195810279-a9a168e8-0a61-4ba6-9831-366f9461d1ae.webm)

We can potentially improve the UI appearance in a follow-up ticket? (as it could be more interesting)

## Check list


samples/app/src/main/kotlin/com/bumble/appyx/app/MainActivity.kt
```diff
@@ -24,14 +26,18 @@ class MainActivity : NodeActivity() {
...
         setContent {
-            AppyxSampleAppTheme {
-                // A surface container using the 'background' color from the theme
-                Surface(color = MaterialTheme.colors.background) {
-                    Column {
-                        NodeHost(integrationPoint = appyxIntegrationPoint) {
-                            RootNode(
-                                buildContext = it,
-                            )
+            CompositionLocalProvider(
```
LachlanMcKee: Needed for the `navigation-compose` sample

zsoltk: I'm a bit unhappy about it as it gives the impression it's needed in general, and people might even copy-paste it needlessly.

LachlanMcKee: I'll see what I can do about it

LachlanMcKee: @zsoltk done. I moved the `LocalIntegrationPoint` to the point at which it is needed by `ComposeNavigationRoot`.

The entire point of that sample is how to use Appyx within an existing Jetpack compose project, so I do not want to expose a root Appyx node from that sample

zsoltk (approved):

Issue #177, opened 2022-09-28
Verify API stability after release


CherryPerry:
For all published modules:

- Use `org.jetbrains.kotlinx:binary-compatibility-validator` and add it as part of CI workflow
- Use `explicitApi()` mode for Kotlin Gradle plugin

PR #230, opened 2022-10-19
API dump

CherryPerry:
## Description

1. Disable generation of `BuildConfig` and etc. to remove them from API (consider as non-breaking, nobody uses them).
2. Generate and save public API dump files. Later these files will be used to verify API stability.
3. Provide guidance in PR template.

Sample from Reaktive repo: https://github.com/badoo/Reaktive/pull/684/files#diff-3072ccd590c25fcd425c2e6496a574be101b854d583830b6e34c4514ede81bd6R392

Part of https://github.com/bumble-tech/appyx/issues/177

## Check list

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -6,3 +6,4 @@
 
 - [ ] I have updated `CHANGELOG.md` if required.
 - [ ] I have updated documentation if required.
+- [ ] I have regenerated API dump via `./gradlew apiDump` task.
--- a/libraries/build.gradle.kts
+++ b/libraries/build.gradle.kts
@@ -1,0 +1,3 @@
+plugins {
+    alias(libs.plugins.kotlin.compatibility)
+}
--- a/gradle.properties
+++ b/gradle.properties
@@ -2,5 +2,9 @@ org.gradle.jvmargs=-Xmx2048m -Dfile.encoding=UTF-8
 org.gradle.caching=true
 org.gradle.parallel=true
 android.useAndroidX=true
+android.defaults.buildfeatures.aidl=false
+android.defaults.buildfeatures.buildconfig=false
+android.defaults.buildfeatures.renderscript=false
+android.defaults.buildfeatures.shaders=false
 kotlin.code.style=official
 library.version=1.0.0-rc01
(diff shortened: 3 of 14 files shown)


.github/pull_request_template.md
```diff
@@ -6,3 +6,4 @@
 
 - [ ] I have updated `CHANGELOG.md` if required.
 - [ ] I have updated documentation if required.
+- [ ] I have regenerated API dump via `./gradlew apiDump` task.
```
LachlanMcKee, line 9: Perhaps as part of CI we could also call `apiDump` and compare it against the PR? We could then fail the build if there is a discrepancy.

That way we don't need this checkbox, and we will avoid the situation where someone regenerates it on their PR and several other PR changes are added into that PR, making reviewing confusing!

libraries/build.gradle.kts
```diff
@@ -0,0 +1,3 @@
+plugins {
```
LachlanMcKee, line 1: interesting! didn't realise this works applies to all the child modules!

LachlanMcKee: Does this then cause this to become a gradle module? (I'm unsure if it was already considered that way before)

.github/pull_request_template.md
CherryPerry: `apiCheck` task is already part of `build` task chain. The problem that I faced in Reaktive, that people do not really check logs and errors, and most of the time we need to tell them explicitly to run this task.

libraries/build.gradle.kts
CherryPerry: `libraries` now is a Gradle module with no tasks but with a plugin.
The plugin is not applied to child modules automatically if it is declared in the parent. The plugin invokes `subprojects { }` internally to apply itself to all subprojects (`libraries:...)`.

.github/pull_request_template.md
LachlanMcKee: They will have to check them if the CI prevents them from merging :D

libraries/build.gradle.kts
LachlanMcKee: thanks for the explanation!

CherryPerry:
We do not want to support binary compatibility, only source one, because we do not expect the library to be used as a transitive dependency.

PR #231, opened 2022-10-19
Add a key to rememberTransitionScope to support dynamically changed transition handlers

KovalevAndrey:
## Description

Changing transition handler at runtime does not result in recomposing children. It happens because `rememberTransitionScope` is remembered and the keys haven't changed. To fix this, I'm adding transition handler instance as a key

Pressing trigger button changes transition handler from fader to slider. To distinguish them, in fader transition handler I added `alpha= 0.5f` for active element

**Before:**

https://user-images.githubusercontent.com/5773436/196752264-0312195f-f820-4d4a-a436-0d74d909a39c.mp4 

**After:**

https://user-images.githubusercontent.com/5773436/196752197-24b8d99e-df41-4326-bd12-c5a7dbb664f8.mp4 

## Check list

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Pending changes
 
+- [#218](https://github.com/bumble-tech/appyx/pull/231) – **Fixed**: Changing transition handler at runtime does not redraw children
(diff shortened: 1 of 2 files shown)


CherryPerry (approved):
Sounds reasonable. We are capturing `this.clipToBoundsModifier` in `remember` and other properties in subclasses, so makes sense to add `this` as a key.

zsoltk (approved):

PR #238, opened 2022-10-20
Sample app update

zsoltk:
## Description

- Rewrote slides around model-driven approach
- Some UI fixes

## Check list


samples/app/src/main/kotlin/com/bumble/appyx/app/node/onboarding/screen/app/StatefulNode2.kt
```diff
@@ -48,8 +48,7 @@ class StatefulNode2(
             body = "– even when they're not visible. " +
                     "\n\nTry going back to the previous screen! " +
                     "You should see that the counters kept on working in the background, " +
-                    "and changes you made to colours are persisted." +
-                    "\n\nIf this behaviour is not desired you can destroy them by using different KeepMode."
```
KovalevAndrey, line 52: no mentioning of destroying `STOPPED` nodes?

KovalevAndrey (approved):

samples/app/src/main/kotlin/com/bumble/appyx/app/node/onboarding/screen/app/StatefulNode2.kt
zsoltk: These slides are not meant to serve as documentation or replace it, but as teasers on large concepts. The removed part should be described in documentation (or on a standalone slide if it's important enough to expose people to it even at the start!). For this text to fit on the same slide, UI proportions needed to be broken initially, I reverted that.

Also I removed these app-tree-related slides altogether, just kept the classes so we can build another deck from them later (but not part of the current model-driven ones).

PR #242, opened 2022-10-21
Remove whats appyx from sample root node

LachlanMcKee:
## Description
Fixed a few minor issues with the sample app.

## Check list


zsoltk:
Worth noting (took me a minute to realise): removing here is ok, since it's already contained in `SamplesContainerNode`

zsoltk (approved):

PR #247, opened 2022-10-23
Fix sample app coroutines onstop

LachlanMcKee:
## Description
Fixed an issue where the sample app samples list screen elements were still active even when it was not active in the backstack.

This is because the `PermanentChild` was actually being called on `SamplesContainerNode` and not `SamplesSelector`. This meant that the children existed outside of the backstack and were never paused/stopped.

Also prevented the teaser NavModels from running when stopped.

## Check list

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Pending changes
 
-–
+– [#247](https://github.com/bumble-tech/appyx/pull/247) – **Added**: Added `EmptyNavModel` to core for cases in which a `ParentNode` only uses `PermanentChild`. The `DummyNavModel` test class is deprecated.
 
 ---
 
--- a/samples/app/src/main/kotlin/com/bumble/appyx/app/node/samples/SamplesSelectorNode.kt
+++ b/samples/app/src/main/kotlin/com/bumble/appyx/app/node/samples/SamplesSelectorNode.kt
@@ -1,0 +1,141 @@
+package com.bumble.appyx.app.node.samples
+
+import android.os.Parcelable
+import androidx.compose.animation.ExperimentalAnimationApi
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.PaddingValues
+import androidx.compose.foundation.layout.aspectRatio
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.lazy.LazyColumn
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.CompositionLocalProvider
+import androidx.compose.runtime.remember
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.ExperimentalComposeUiApi
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.unit.ExperimentalUnitApi
+import androidx.compose.ui.unit.dp
+import com.bumble.appyx.app.node.cards.CardsExampleNode
+import com.bumble.appyx.app.node.slideshow.WhatsAppyxSlideShow
+import com.bumble.appyx.core.composable.ChildRenderer
+import com.bumble.appyx.core.integrationpoint.LocalIntegrationPoint
+import com.bumble.appyx.core.modality.BuildContext
+import com.bumble.appyx.core.navigation.EmptyNavModel
+import com.bumble.appyx.core.node.Node
+import com.bumble.appyx.core.node.ParentNode
+import com.bumble.appyx.core.node.node
+import com.bumble.appyx.navmodel.backstack.BackStack
+import com.bumble.appyx.navmodel.backstack.operation.push
+import com.bumble.appyx.sample.navigtion.compose.ComposeNavigationRoot
+import kotlinx.parcelize.Parcelize
+
+class SamplesSelectorNode(
+    buildContext: BuildContext,
+    private val samplesContainerBackStack: BackStack<SamplesContainerNode.NavTarget>
+) : ParentNode<SamplesSelectorNode.NavTarget>(
+    navModel = EmptyNavModel<NavTarget, Unit>(),
+    buildContext = buildContext
+) {
+    sealed class NavTarget : Parcelable {
+        @Parcelize
+        object OnboardingScreen : NavTarget()
+
+        @Parcelize
+        object ComposeNavigationScreen : NavTarget()
+
+        @Parcelize
+        object CardsExample : NavTarget()
+    }
+
+    @ExperimentalUnitApi
+    @ExperimentalAnimationApi
+    @ExperimentalComposeUiApi
+    override fun resolve(navTarget: NavTarget, buildContext: BuildContext): Node =
+        when (navTarget) {
+            is NavTarget.CardsExample -> CardsExampleNode(buildContext)
+            is NavTarget.OnboardingScreen -> WhatsAppyxSlideShow(
+                buildContext = buildContext,
+                autoAdvanceDelayMs = 2500
+            )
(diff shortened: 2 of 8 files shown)


LachlanMcKee:
We could potentially wait for 1.1.0? This only impacts Samples, but I think the NoOpNavModel class has reason to exist in the core API based on this usecase

CHANGELOG.md
```diff
@@ -2,7 +2,8 @@
 
 ## Pending changes
 
-–
+– [#247](https://github.com/bumble-tech/appyx/pull/247) – **Breaking change**: Removed `DummyNavModel` from testing libraries. Use `NoOpNavModel` instead.
```
KovalevAndrey: let's not introduce any breaking api changes for 1.0

samples/app/src/main/kotlin/com/bumble/appyx/app/node/samples/SamplesSelectorNode.kt
```diff
@@ -0,0 +1,159 @@
...
+import com.bumble.appyx.core.node.ParentNode
+import com.bumble.appyx.core.node.node
+import com.bumble.appyx.navmodel.backstack.BackStack
+import com.bumble.appyx.navmodel.backstack.operation.push
+import com.bumble.appyx.sample.navigtion.compose.ComposeNavigationRoot
+import kotlinx.parcelize.Parcelize
+
+class SamplesSelectorNode(
+    buildContext: BuildContext,
+    private val samplesContainerBackStack: BackStack<SamplesContainerNode.NavTarget>
```
KovalevAndrey: children have dependency on the parent, this is bad practice, they should produce an output instead

samples/app/src/main/kotlin/com/bumble/appyx/app/node/samples/SamplesSelectorNode.kt
```diff
@@ -0,0 +1,159 @@
...
+
+    private fun LazyListScope.cardsItem(
+        decorator: @Composable (child: ChildRenderer) -> Unit
+    ) {
+        item {
+            SampleItem(
+                title = "Dating cards NavModel",
+                subtitle = "Swipe right on the NavModel concept",
+                onClick = {
+                    samplesContainerBackStack.push(SamplesContainerNode.NavTarget.CardsExample)
```
KovalevAndrey: let's produce an output, and parent will preform actions on the routing source

KovalevAndrey (changes requested):

CHANGELOG.md
LachlanMcKee: what do you propose? wait for 1.0 to be cut? or move this model to samples for now with a plan to move it to core later?

samples/app/src/main/kotlin/com/bumble/appyx/app/node/samples/SamplesSelectorNode.kt
LachlanMcKee: can do :)

CHANGELOG.md
CherryPerry: You do not need to remove it. Introduce type alias + deprecate notice.

libraries/core/src/main/kotlin/com/bumble/appyx/core/navigation/NoOpNavModel.kt
```diff
@@ -0,0 +1,17 @@
+package com.bumble.appyx.core.navigation
+
+import com.bumble.appyx.core.navigation.onscreen.OnScreenStateResolver
+
+/**
+ * An implementation of a NavModel that won't add any children.
+ * This is potentially useful if your ParentNode only uses
+ * [com.bumble.appyx.core.navigation.model.permanent.PermanentNavModel]
+ */
+class NoOpNavModel<NavTarget, State> : BaseNavModel<NavTarget, State>(
```
CherryPerry: Would prefer `EmptyNavModel`. I think will be easier to find naturally via autocomplete.

CHANGELOG.md
LachlanMcKee: yep, will do!

libraries/core/src/main/kotlin/com/bumble/appyx/core/navigation/NoOpNavModel.kt
LachlanMcKee: done

samples/app/src/main/kotlin/com/bumble/appyx/app/node/samples/SamplesSelectorNode.kt
LachlanMcKee: @KovalevAndrey done. though should I use a NodeConnector? It doesn't necessarily need to be done with Rx, and I'm not quite sure what our approach is when we don't use Rx

samples/app/src/main/kotlin/com/bumble/appyx/app/node/teaser/backstack/BackstackTeaserNode.kt
```diff
@@ -36,18 +36,20 @@ class BackstackTeaserNode(
 
     init {
         lifecycle.coroutineScope.launch {
-            delay(1000)
-            repeat(4) {
-                backStack.push(NavTarget.Child((it + 2) * 100))
-                delay(400)
+            lifecycle.coroutineScope.launchWhenStarted {
```
CherryPerry: Why do we need to use both `launch` and `launchWhenStarted`? Just `launchWhenStarted` should be enough, it is not suspending.

samples/app/src/main/kotlin/com/bumble/appyx/app/node/teaser/promoter/PromoterTeaserNode.kt
```diff
@@ -37,17 +37,19 @@ class PromoterTeaserNode(
 
     init {
         lifecycle.coroutineScope.launch {
-            repeat(4) {
-                promoter.addFirst(NavTarget.Child((it + 1) * 100))
-                promoter.promoteAll()
+            lifecycle.coroutineScope.launchWhenStarted {
```
CherryPerry: Same here.

CherryPerry (approved):

samples/app/src/main/kotlin/com/bumble/appyx/app/node/teaser/backstack/BackstackTeaserNode.kt
LachlanMcKee: This is the recommend approach from Google, the other approach is deprecated.

This code also ensures that it stops the coroutine when I stop happens

samples/app/src/main/kotlin/com/bumble/appyx/app/node/teaser/promoter/PromoterTeaserNode.kt
LachlanMcKee: Same as above

KovalevAndrey (approved):

samples/app/src/main/kotlin/com/bumble/appyx/app/node/teaser/backstack/BackstackTeaserNode.kt
LachlanMcKee: Oh actually, I'll double check this, I meant to use a different method. I might revert this for this PR, and add the other change later

LachlanMcKee: Fixed.

Issue #234, opened 2022-10-20
Fix double splash screen in the main sample app


zsoltk:
We need to migrate our splash screen in `:samples:app` to avoid a duplicated effect. We already use an internal one, see `RootNode`, `NavTarget.Splash`.

https://developer.android.com/develop/ui/views/launch/splash-screen/migrate

PR #248, opened 2022-10-24
Remove custom splash and use new Android API for Splash screen

mapm14:
## Description

https://github.com/bumble-tech/appyx/issues/234

**Dark mode:**

**Light mode:**

## Check list


samples/app/src/main/res/values/themes.xml
```diff
@@ -10,12 +10,21 @@
...
     </style>
 
     <style name="Theme.Appyx.NoActionBar">
         <item name="windowActionBar">false</item>
         <item name="windowNoTitle">true</item>
     </style>
+
+    <style name="Theme.Appyx.Starting" parent="Theme.SplashScreen">
+        <!-- Set the splash screen background -->
+        <item name="windowSplashScreenBackground">@color/appyx_dark</item>
```
LachlanMcKee: How does this work with dark/light theme like we do in the node at the moment?

samples/app/src/main/kotlin/com/bumble/appyx/app/MainActivity.kt
```diff
@@ -22,14 +23,15 @@ import com.bumble.appyx.core.modality.BuildContext
 class MainActivity : NodeActivity() {
 
     override fun onCreate(savedInstanceState: Bundle?) {
+        installSplashScreen()
```
LachlanMcKee, line 26: Should we use `installSplashScreen()`? I believe this is for legacy projects where you already have a splash screen. Is there not a dedicated splash screen activity in the library?

LachlanMcKee: https://developer.android.com/develop/ui/views/launch/splash-screen/migrate#remove_the_custom_splash_screen_activity

mapm14: If I don't do it the app crashes (at least on my device)
It's part of the step number 4 in [here](https://developer.android.com/develop/ui/views/launch/splash-screen/migrate#migrate_your_splash_screen_implementation)

samples/app/src/main/res/values/themes.xml
mapm14: Good catch! I forgot to modify the theme for dark mode, is done now.

samples/app/src/main/kotlin/com/bumble/appyx/app/MainActivity.kt
LachlanMcKee: ah fair enough. I thought the API suggested that they had their own activity, and for developers to not use their own activity for this purpose.

LachlanMcKee (approved):

PR #249, opened 2022-10-24
Remove dependency analysis from ci reports

LachlanMcKee:
## Description
Removed the dependency analysis reports from CI as they are not useful (at the moment) and increase the file size of the ZIP from roughly 3MB to 200MB

-------
**Failure job with dependency analysis**
https://github.com/bumble-tech/appyx/actions/runs/3312319396

image
-------
**Failure job without dependency analysis**
https://github.com/bumble-tech/appyx/actions/runs/3312440201

image
-------

## Check list

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -68,7 +68,9 @@ jobs:
         uses: actions/upload-artifact@v3
         with:
           name: reports
-          path: '**/build/reports'
+          path: |
+            **/build/reports/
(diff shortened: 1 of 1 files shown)


LachlanMcKee:
One of the consequences of this is that downloading the report can actually take quite a while. It seems that github doesn't necessarily provide great bandwidth when downloading artifacts, so having them smaller is quite important.

CherryPerry (approved):

PR #251, opened 2022-10-25
Update icon to svg + clipping dating cards

KovalevAndrey:
## Description

Updated icon to svg and used clip to bounds for dating cards

## Check list


zsoltk (approved):

samples/app/src/main/res/drawable/appyx.xml
```diff
@@ -0,0 +1,37 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
LachlanMcKee, line 1: @KovalevAndrey does this work on lollipop? Or should we change our min SDK for the samples app to a version that supports this without crashing?

PR #252, opened 2022-10-25
Don't handle backpress for previews

KovalevAndrey:
## Description

Don't handle back press for preview in What's appyx slider Node

## Check list


samples/app/src/main/kotlin/com/bumble/appyx/app/node/samples/SamplesContainerNode.kt
```diff
@@ -163,8 +167,22 @@ class SamplesContainerNode(
                     modifier = Modifier
                         .fillMaxSize()
                         .aspectRatio(16f / 9),
-                    onClick = { backStack.push(NavTarget.OnboardingScreen(false)) },
-                ) { PermanentChild(navTarget = NavTarget.OnboardingScreen(true), decorator = decorator) }
+                    onClick = {
+                        backStack.push(
+                            NavTarget.OnboardingScreen(
+                                false,
```
mapm14: Can we please add names params for this variable? Also for line 181

mapm14 (approved):

samples/app/src/main/kotlin/com/bumble/appyx/app/node/slideshow/WhatsAppyxSlideShow.kt
```diff
@@ -64,14 +65,19 @@ import kotlinx.parcelize.Parcelize
 class WhatsAppyxSlideShow(
     buildContext: BuildContext,
     autoAdvanceDelayMs: Long? = null,
+    private val isInPreviewMode: Boolean = false,
```
zsoltk: These two params belong together. If we're in preview, we also have an `autoAdvanceDelayMs`. Maybe express it with type and package it into a `LaunchMode` sealed class. (Separate PR ok too to unblock this as a quick fix).

zsoltk (approved):

LachlanMcKee (approved):

PR #255, opened 2022-10-25
Publish sample app

LachlanMcKee:
## Description
Publish sample app and link it on the documentation.
In a follow up PR I will look into uploading this to the release as well.

## Check list


.github/workflows/post_merge.yml
```diff
@@ -0,0 +1,39 @@
...
+    name: Build sample app
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-java@v3
+        with:
+          distribution: 'zulu'
+          java-version: '11'
+      - uses: gradle/wrapper-validation-action@v1
+      - uses: gradle/gradle-build-action@v2
```
CherryPerry, line 32: What do you think about `cache-read-only: true` here?
Because changes in both core and samples will trigger cache invalidation, does not really makes sense to keep it.

CherryPerry (approved):

.github/workflows/post_merge.yml
LachlanMcKee: Sure. I can add it

PR #266, opened 2022-11-02
Example of a tree navigation

### LINKED ISSUE #229: Provide Navigator instance to Appyx tree to able to execute workflows inside Nodes
Currently we can execute workflow inside activity when, for instance, handling a deeplink. We should provide a mechanism for workflow executions inside Node of the Appyx tree.

KovalevAndrey:
## Description

Fixes https://github.com/bumble-tech/appyx/issues/229

Example of how you can implement a global tree navigator to use to navigate from inside the Appyx tree

https://user-images.githubusercontent.com/5773436/199481939-1b9907c0-1826-46d0-a0b5-434b29de1fff.mp4

## Check list

### DIFF
--- a/samples/sandbox/src/main/kotlin/com/bumble/appyx/sandbox/client/workflow/RootNode.kt
+++ b/samples/sandbox/src/main/kotlin/com/bumble/appyx/sandbox/client/workflow/RootNode.kt
@@ -5,11 +5,13 @@ import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.requiredHeight
 import androidx.compose.material.Button
 import androidx.compose.material.Text
 import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment.Companion.CenterHorizontally
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.unit.dp
 import com.bumble.appyx.core.composable.Children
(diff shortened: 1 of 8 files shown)


samples/sandbox/src/main/kotlin/com/bumble/appyx/sandbox/client/workflow/WorkflowExampleActivity.kt
```diff
@@ -75,21 +80,37 @@ class WorkflowExampleActivity : NodeActivity() {
...
     }
 
-    private fun executeWorkflow1(): Job {
-        return lifecycleScope.launch {
+    private fun executeNavigation(navigationAction: suspend () -> Unit) {
+        job?.cancel()
+        job = lifecycleScope.launch {
+            navigationAction()
+        }
+    }
```
zsoltk, line 88: I feel this should be part of `executeWorkflow` internally in `Node`. Not just to reduce code here, but the lifecycle scoping should belong to the individual nodes rather than the entire activity.

samples/sandbox/src/main/kotlin/com/bumble/appyx/sandbox/client/workflow/RootNode.kt
```diff
@@ -41,9 +46,9 @@ class RootNode(
         return waitForChildAttached()
     }
 
-    suspend fun attachChildTwo(): ChildNodeTwo {
+    suspend fun attachChildOne(): ChildNodeOne {
         return attachWorkflow {
```
zsoltk: Expression syntax maybe?

samples/sandbox/src/main/kotlin/com/bumble/appyx/sandbox/client/workflow/WorkflowExampleActivity.kt
KovalevAndrey: > I feel this should be part of `executeWorkflow` internally in `Node`. Not just to reduce code here, but the lifecycle scoping should belong to the individual nodes rather than the entire activity.

This this the global scope. It will be cancelled if activity gets `onDestoy` callback. Individual nodes use their own `lifecycleScope` so they will be cancelled when the node is destoyed

zsoltk (approved):

PR #268, opened 2022-11-08
Fix PermanentChild behaviour in test environment

CherryPerry:
## Description

`PermanentChild` throws exception in UI tests due unsynchronized changes from `collect`.

Added new `PermanentChild.addUnique` API to use from `PermanentChild`.

## Check list

### DIFF
--- a/libraries/core/src/main/kotlin/com/bumble/appyx/core/navigation/model/permanent/operation/AddUnique.kt
+++ b/libraries/core/src/main/kotlin/com/bumble/appyx/core/navigation/model/permanent/operation/AddUnique.kt
@@ -7,6 +7,9 @@ import com.bumble.appyx.core.navigation.model.permanent.PermanentNavModel
 import kotlinx.parcelize.Parcelize
 import kotlinx.parcelize.RawValue
 
+/**
+ * Adds [NavTarget] into [PermanentNavModel] only if it is not there yet.
+ */
 @Parcelize
 data class AddUnique<NavTarget : Any>(
     private val navTarget: @RawValue NavTarget,
--- a/libraries/core/src/main/kotlin/com/bumble/appyx/core/navigation/model/permanent/operation/AddUnique.kt
+++ b/libraries/core/src/main/kotlin/com/bumble/appyx/core/navigation/model/permanent/operation/AddUnique.kt
@@ -1,0 +1,35 @@
+package com.bumble.appyx.core.navigation.model.permanent.operation
+
+import com.bumble.appyx.core.navigation.NavElement
+import com.bumble.appyx.core.navigation.NavElements
+import com.bumble.appyx.core.navigation.NavKey
+import com.bumble.appyx.core.navigation.model.permanent.PermanentNavModel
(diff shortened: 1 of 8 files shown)


libraries/core/src/main/kotlin/com/bumble/appyx/core/navigation/model/permanent/operation/AddUnique.kt
```diff
@@ -0,0 +1,35 @@
...
+
+import com.bumble.appyx.core.navigation.NavElement
+import com.bumble.appyx.core.navigation.NavElements
+import com.bumble.appyx.core.navigation.NavKey
+import com.bumble.appyx.core.navigation.model.permanent.PermanentNavModel
+import kotlinx.parcelize.Parcelize
+import kotlinx.parcelize.RawValue
+
+@Parcelize
+data class AddUnique<NavTarget : Any>(
```
LachlanMcKee, line 14: Is this expected to be called by library users? otherwise internal? if it should be public, add some Javadoc? Maybe add some either way?

libraries/core/src/main/kotlin/com/bumble/appyx/core/node/ParentNode.kt
```diff
@@ -98,15 +99,19 @@ abstract class ParentNode<NavTarget : Any>(
...
-                    ?: NavKey(navTarget).also { permanentNavModel.add(it) }
-                child = childOrCreate(navKey)
-            }
+            permanentNavModel.addUnique(navTarget)
         }
-
+        val child by remember(navTarget) {
+            permanentNavModel
+                .elements
+                .map { navElements ->
```
LachlanMcKee: mapState? Also, StateFlows are already 'distinctUntilChanged', so you don't need that operator if you do this.

libraries/core/src/main/kotlin/com/bumble/appyx/core/node/ParentNode.kt
```diff
@@ -98,15 +99,19 @@ abstract class ParentNode<NavTarget : Any>(
...
+                    navElements
+                        .find { it.key.navTarget == navTarget }
+                        ?.let { childOrCreate(it.key) }
+                }
+                .distinctUntilChanged()
+        }.collectAsState(initial = null)
```
LachlanMcKee: If you `mapState` you don't need initial

libraries/core/src/main/kotlin/com/bumble/appyx/core/navigation/model/permanent/operation/AddUnique.kt
CherryPerry: I think it is more useful than regular `Add`.

libraries/core/src/main/kotlin/com/bumble/appyx/core/node/ParentNode.kt
CherryPerry: `StateFlows` are, but `map` result is not.

LachlanMcKee (approved):

PR #272, opened 2022-11-15
Deep linking navigation

KovalevAndrey:
## Description

Fixes https://github.com/bumble-tech/appyx/issues/87

Documentation about explicit navigation and deep linking

## Check list

- [x] I have updated `CHANGELOG.md` if required.
- [x] I have updated documentation if required.

PR #274, opened 2022-11-15
Avoid breaking changes by reinstantiating and deprecating API

KovalevAndrey:
## Description

Deprecate renamed API instead of deleting it to avoid breaking changes

## Check list

### DIFF
--- a/libraries/core/src/main/kotlin/com/bumble/appyx/core/node/Node.kt
+++ b/libraries/core/src/main/kotlin/com/bumble/appyx/core/node/Node.kt
@@ -100,6 +100,11 @@ open class Node(
         )
     }
 
+    @Deprecated("Will be removed in 1.1. Use executeAction instead")
+    protected suspend inline fun <reified T : Node> executeWorkflow(
+        crossinline action: () -> Unit
+    ): T = executeAction(action)
+
     protected suspend inline fun <reified T : Node> executeAction(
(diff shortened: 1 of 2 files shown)


libraries/core/src/main/kotlin/com/bumble/appyx/core/node/Node.kt
```diff
@@ -100,6 +100,11 @@ open class Node(
         )
     }
 
+    @Deprecated("Will be removed in 1.1. Use executeAction instead")
```
zsoltk: Let's add `replaceWith`

zsoltk (approved):

libraries/core/src/main/kotlin/com/bumble/appyx/core/node/Node.kt
CherryPerry: You can add `replaceWith = ReplaceWith("executeAction(action)")` so IDEA can replace it automatically.

LachlanMcKee:
Maybe call this out in the changelog?

KovalevAndrey:
> Maybe call this out in the changelog?

already there https://github.com/bumble-tech/appyx/pull/272/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR6

PR #276, opened 2022-11-18
Fix back press invocation order in RIBs-Appyx integration

CherryPerry:
### Issue:

1. There is RIBs container with a backstack that can host the main page and the create event page (appyx).
2. We navigate from the main to the create event page.
3. RIBs backstack has 2 elements and can go back.
4. The create event page want to intercept back press and registers back press listener via AndroidX API.
5. When I press click it goes to RIBs first.
6. The main screen RIB tries to pass the back click event to the children first (RIBs-Appyx Interop node) via RIBs back press API.
7. InteropNode does not know how to handle back press so it ignores it.
8. The back click now is handled by the main screen RIB router so we navigate back.
9. RIB returns that the back click is handled so super<Activity>.onBackPressed() is not invoked leading to skip of AndroidX API.

### Solution

Provide custom `OnBackPressedDispatcher` for every RIBs-Appyx integration point via `CompositionLocalProvider`. Appyx will use it to register all callbacks. Then we manually invoke them from RIBs `BackPressHandler` plugin.

## Check list


libraries/interop-ribs/src/main/kotlin/com/bumble/appyx/interop/ribs/InteropBackPressHandler.kt
```diff
@@ -0,0 +1,45 @@
...
+ * 1. All active RIBs handlers using RIBs back press API.
+ * 2. All active Appyx handlers using AndroidX back press API.
+ *
+ * The reason is RIBs integration point implementation
+ * where `super<Activity>.onBackPressed()` is invoked after `Node.handleBackPress()`.
+ *
+ * Let's collect all Appyx back press handlers into a separate [OnBackPressedDispatcher]
+ * and use it for `handleBackPress()`.
+ */
+internal class InteropBackPressHandler :
```
KovalevAndrey, line 20: should we create a quick sample for other to refer to?

KovalevAndrey (approved):

PR #287, opened 2022-11-28
Updated twitter compose detekt to 0.0.26

LachlanMcKee:
## Description
Updated twitter compose detekt to 0.0.26

## Check list

### DIFF
--- a/libraries/core/src/main/kotlin/com/bumble/appyx/core/collections/ImmutableList.kt
+++ b/libraries/core/src/main/kotlin/com/bumble/appyx/core/collections/ImmutableList.kt
@@ -1,0 +1,12 @@
+package com.bumble.appyx.core.collections
+
+import androidx.compose.runtime.Immutable
+
+@Immutable
+data class ImmutableList<out E>(private val list: List<E>) : List<E> by list
+
+fun <E> immutableListOf(vararg items: E): ImmutableList<E> =
+    ImmutableList(listOf(*items))
+
+fun <E> List<E>.toImmutableList(): ImmutableList<E> =
+    ImmutableList(ArrayList(this)) // make a copy as list might be also mutable
--- a/libraries/core/src/main/kotlin/com/bumble/appyx/core/navigation/transition/CombinedHandler.kt
+++ b/libraries/core/src/main/kotlin/com/bumble/appyx/core/navigation/transition/CombinedHandler.kt
@@ -5,6 +5,7 @@ import androidx.compose.animation.core.Transition
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.remember
 import androidx.compose.ui.Modifier
+import com.bumble.appyx.core.collections.ImmutableList
 
 class CombinedHandler<T, S>(
     private val handlers: List<ModifierTransitionHandler<T, S>>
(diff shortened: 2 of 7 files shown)


libraries/core/build.gradle.kts
```diff
@@ -38,6 +38,7 @@ dependencies {
     api(libs.compose.animation.core)
     api(libs.compose.runtime)
     api(libs.androidx.appcompat)
+    api(libs.kotlinx.collections.immutable)
```
KovalevAndrey: Last time we discussed it with @CherryPerry we decided not to use it. Latest release was around a [year ago](https://github.com/Kotlin/kotlinx.collections.immutable/tags) and there hasn't been any activity for a while

LachlanMcKee: @KovalevAndrey heh. I was going to discuss this later today. you beat me to it!
We might need to 'roll our own' immutable list data structure then?

I kept it as a draft due to being unsure about what approach to take.

LachlanMcKee: @KovalevAndrey I have removed the dependency and added our own implementation of immutable list. The PR is also now ready for review!

libraries/core/src/main/kotlin/com/bumble/appyx/core/collections/ImmutableList.kt
```diff
@@ -0,0 +1,12 @@
+package com.bumble.appyx.core.collections
+
+import androidx.compose.runtime.Immutable
+
+@Immutable
+data class ImmutableList<out E>(val list: List<E>) : List<E> by list
```
KovalevAndrey: can be private? otherwise you can modify list like `(immutableList.list as MutableList<E>).add(E)` if you wrapped around mutable list
```suggestion
data class ImmutableList<out E>(private val list: List<E>) : List<E> by list
```

libraries/core/src/main/kotlin/com/bumble/appyx/core/navigation/transition/CombinedHandler.kt
```diff
@@ -24,5 +25,7 @@ class CombinedHandler<T, S>(
 }
 
 @Composable
-fun <T, S> rememberCombinedHandler(handlers: List<ModifierTransitionHandler<T, S>>): ModifierTransitionHandler<T, S> =
+fun <T, S> rememberCombinedHandler(
+    handlers: ImmutableList<ModifierTransitionHandler<T, S>>
+): ModifierTransitionHandler<T, S> =
     remember { CombinedHandler(handlers = handlers) }
```
KovalevAndrey:
```suggestion
    remember(handlers) { CombinedHandler(handlers = handlers) }
```

LachlanMcKee: That's scope creep, though I guess this is already a slight API change =P

libraries/core/src/main/kotlin/com/bumble/appyx/core/collections/ImmutableList.kt
LachlanMcKee: done

libraries/core/src/main/kotlin/com/bumble/appyx/core/navigation/transition/CombinedHandler.kt
LachlanMcKee: done

libraries/core/src/main/kotlin/com/bumble/appyx/core/navigation/transition/CombinedHandler.kt
```diff
@@ -24,5 +25,7 @@ class CombinedHandler<T, S>(
 }
 
 @Composable
-fun <T, S> rememberCombinedHandler(handlers: List<ModifierTransitionHandler<T, S>>): ModifierTransitionHandler<T, S> =
-    remember { CombinedHandler(handlers = handlers) }
+fun <T, S> rememberCombinedHandler(
```
KovalevAndrey, line 40: Let's deprecate the current `rememberCombinedHandler` and introduce a new one that takes `ImmutableList` in order not to break API, and remove it in 1.1?

LachlanMcKee: Good idea. Will do

LachlanMcKee: @KovalevAndrey done

KovalevAndrey (approved):

PR #297, opened 2022-12-04
Issue #296 Dispose NodeConnector intake in onDestroy

### LINKED ISSUE #296: NodeConnector intake is not disposed when a node is destroyed
It seems that not disposing `NodeConnector.intake` when a Node is destroyed is a bug.

Please let me know if this is intended. It could also potentially be a breaking change, so perhaps we need to throw an error if it happens

LachlanMcKee:
## Description
Closes #296 

## Check list

### DIFF
--- a/libraries/interop-rx2/src/main/kotlin/com/bumble/appyx/interop/rx2/connectable/NodeConnector.kt
+++ b/libraries/interop-rx2/src/main/kotlin/com/bumble/appyx/interop/rx2/connectable/NodeConnector.kt
@@ -5,6 +5,7 @@ import com.bumble.appyx.core.lifecycle.subscribe
 import com.jakewharton.rxrelay2.PublishRelay
 import com.jakewharton.rxrelay2.Relay
 import io.reactivex.Observer
+import io.reactivex.disposables.Disposable
 
 class NodeConnector<Input, Output>(
     override val input: Relay<Input> = PublishRelay.create(),
@@ -15,6 +16,7 @@ class NodeConnector<Input, Output>(
     private val exhaust: Relay<Output> = PublishRelay.create()
     private var isFlushed = false
     private val outputCache = mutableListOf<Output>()
+    private var intakeDisposable: Disposable? = null
 
     override val output: Relay<Output> = object : Relay<Output>() {
 
@@ -31,7 +33,7 @@ class NodeConnector<Input, Output>(
     }
 
     override fun onCreate(lifecycle: Lifecycle) {
-        lifecycle.subscribe(onCreate = { flushOutputCache() })
+        lifecycle.subscribe(onCreate = { flushOutputCache() }, onDestroy = { intakeDisposable?.dispose() })
     }
 
     private val cacheSubscription = intake.subscribe {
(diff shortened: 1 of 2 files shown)


libraries/interop-rx2/src/main/kotlin/com/bumble/appyx/interop/rx2/connectable/NodeConnector.kt
```diff
@@ -31,7 +33,7 @@ class NodeConnector<Input, Output>(
     }
 
     override fun onCreate(lifecycle: Lifecycle) {
-        lifecycle.subscribe(onCreate = { flushOutputCache() })
+        lifecycle.subscribe(onCreate = { flushOutputCache() }, onDestroy = { intakeDisposable?.dispose() })
```
LachlanMcKee, line 36: Possibly we should log an error, or raise an exception if input/output is invoked after `onDestroy` is called, rather than just swallowing silently?

libraries/interop-rx2/src/main/kotlin/com/bumble/appyx/interop/rx2/connectable/NodeConnector.kt
```diff
@@ -55,7 +57,7 @@ class NodeConnector<Input, Output>(
     }
 
     private fun switchToExhaust() {
-        intake.subscribe { exhaust.accept(it) }
+        intakeDisposable = intake.subscribe { exhaust.accept(it) }
```
CherryPerry, line 60: `intake.subscribe(exhaust)`

CherryPerry:
Honestly do not think it is an error. Both source and subscriber are managed by us, they should be GCed together as soon as everybody unsubscribed from node.

But not against PR, makes sense.

CherryPerry (approved):

libraries/interop-rx2/src/main/kotlin/com/bumble/appyx/interop/rx2/connectable/NodeConnector.kt
CherryPerry: You can't skip `onCreate` callback. `androidx.lifecycle.LifecycleRegistry.moveToState` forbids going from initialised state to destroyed.

LachlanMcKee: sorry, I wasn't talking about lifecycles here @CherryPerry. I meant if someone called `input.accept`/`output.accept` after onDestroy for the node was called.

In this case the output event would be ignored, but the input event would not.

libraries/interop-rx2/src/main/kotlin/com/bumble/appyx/interop/rx2/connectable/NodeConnector.kt
LachlanMcKee: makes sense

LachlanMcKee:
@CherryPerry the garbage collector is not deterministic, so if we don't unsubscribe in a deterministic manner, users may get unexpected behaviours (depending on whether they are still sending events after a node is destroyed.

libraries/interop-rx2/src/main/kotlin/com/bumble/appyx/interop/rx2/connectable/NodeConnector.kt
CherryPerry: IMHO do not see it as a problem that we should take care of, swallowing is OK for me.

PR #300, opened 2022-12-07
Updated espresso-core to 3.5.0

LachlanMcKee:
## Description
Updated espresso-core to 3.5.0 (as this is a test library, we don't need to update the change log)

## Check list

### DIFF
--- a/libraries/core/src/androidTest/kotlin/com/bumble/appyx/core/plugin/BackPressHandlerTest.kt
+++ b/libraries/core/src/androidTest/kotlin/com/bumble/appyx/core/plugin/BackPressHandlerTest.kt
@@ -1,6 +1,5 @@
 package com.bumble.appyx.core.plugin
 
-import android.app.Activity
 import android.os.Parcelable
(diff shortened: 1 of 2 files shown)


libraries/core/src/androidTest/kotlin/com/bumble/appyx/core/plugin/BackPressHandlerTest.kt
```diff
@@ -91,7 +91,7 @@ class BackPressHandlerTest {
         Espresso.onIdle()
         rule.waitForIdle()
 
-        assertThat(rule.activityScenario.result.resultCode, equalTo(Activity.RESULT_CANCELED))
```
LachlanMcKee, line 94: This test was failing with the following exception: `You must start Activity first. Make sure you are using launchActivityForResult() to launch an Activity.`.

Our tests are not calling `launchActivityForResult()`, hence why the result code no longer works. I believe checking that the activityScenario state is destroyed should be sufficient.

CherryPerry (approved):

PR #303, opened 2022-12-13
ViewModel POC

KovalevAndrey:
## Description

This PR demonstrates how VM can be used and scoped to a `Node` using reflection API.

By design VM is stored in a `ViewModelStore` class and lives as long as activity lives. 

The desired behaviour is the following:
1. `VM` is created when `Node` is created
2. `VM` is persistent when `Node` is moved in `DESTROY` state and `Activity` is changing configuration
3. `VM` is destroyed when `Node` is moved in `DESTROY` state and `Activity` **IS NOT** changing configuration

This behaviour can not be achieved using public API as it's not possible to clear only one `VM` from `ViewModelStore. It has only one public method – `ViewModelStore.clear()` which clears all `VMs.

I the meantime I'm exploring alternative solutions.

https://user-images.githubusercontent.com/5773436/207447633-65f79414-2313-4d7b-b453-464c4e51de30.mp4

## Check list

### DIFF
--- a/samples/sandbox/src/main/kotlin/com/bumble/appyx/sandbox/client/viewmodel/ViewModelNode.kt
+++ b/samples/sandbox/src/main/kotlin/com/bumble/appyx/sandbox/client/viewmodel/ViewModelNode.kt
@@ -1,0 +1,48 @@
+package com.bumble.appyx.sandbox.client.viewmodel
+
+import androidx.activity.viewModels
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.material.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.collectAsState
+import androidx.compose.runtime.getValue
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.unit.sp
+import androidx.lifecycle.DefaultLifecycleObserver
+import androidx.lifecycle.LifecycleOwner
+import com.bumble.appyx.core.integrationpoint.ActivityIntegrationPoint
+import com.bumble.appyx.core.modality.BuildContext
+import com.bumble.appyx.core.node.Node
+
+class ViewModelNode(
+    buildContext: BuildContext,
+) : Node(
+    buildContext = buildContext
+) {
+    private val activity = (integrationPoint as ActivityIntegrationPoint).activity
+    private val viewModel: MyViewModel by activity.viewModels { MyViewModel.Factory }
+
+    init {
+        lifecycle.addObserver(object : DefaultLifecycleObserver {
+            override fun onDestroy(owner: LifecycleOwner) {
+                if (!activity.isChangingConfigurations) {
+                    activity.viewModelStore.removeViewModel(viewModel.javaClass)
+                }
+            }
+        })
+    }
+
+    @Composable
+    override fun View(modifier: Modifier) {
+        val counter by viewModel.flow.collectAsState(initial = 0)
+        Box(modifier = modifier.fillMaxSize()) {
+            Text(
+                text = counter.toString(),
+                fontSize = 38.sp,
+                modifier = Modifier.align(Alignment.Center)
+            )
+        }
+    }
+}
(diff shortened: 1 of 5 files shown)


samples/sandbox/src/main/kotlin/com/bumble/appyx/sandbox/client/viewmodel/ViewModelNode.kt
```diff
@@ -0,0 +1,48 @@
...
+import com.bumble.appyx.core.integrationpoint.ActivityIntegrationPoint
+import com.bumble.appyx.core.modality.BuildContext
+import com.bumble.appyx.core.node.Node
+
+class ViewModelNode(
+    buildContext: BuildContext,
+) : Node(
+    buildContext = buildContext
+) {
+    private val activity = (integrationPoint as ActivityIntegrationPoint).activity
```
KovalevAndrey, line 24: I can wrap all required code in a less uglier way exposing only VM factory

CherryPerry:
IMHO it should be implemented in the same way as in `Fragment` – each `Node` should be `ViewModelStoreOwner` with own `ViewModelStore`. Then throw everything into single `FragmentManagerViewModel`-like `ViewModel` that we will register once in activity integration point.

Anyway in both variants we have another problem – we change DI flow. We usually pass all dependencies into `Node` ready to use, but here we have to instantiate `ViewModel` inside `Node`, comparing to everything else that is done outside. Not like a bad thing, something that I want to come to later.

RationalRank:
> IMHO it should be implemented in the same way as in `Fragment` – each `Node` should be `ViewModelStoreOwner` with own `ViewModelStore`. Then throw everything into single `FragmentManagerViewModel`-like `ViewModel` that we will register once in activity integration point.
> 
> Anyway in both variants we have another problem – we change DI flow. We usually pass all dependencies into `Node` ready to use, but here we have to instantiate `ViewModel` inside `Node`, comparing to everything else that is done outside. Not like a bad thing, something that I want to come to later.

This is what we're doing internally right now. I've published a [sample](https://github.com/RationalRank/appyx_viewmodel_poc) demonstrating this

KovalevAndrey:
> IMHO it should be implemented in the same way as in `Fragment` – each `Node` should be `ViewModelStoreOwner` with own `ViewModelStore`. Then throw everything into single `FragmentManagerViewModel`-like `ViewModel` that we will register once in activity integration point.
> 
> Anyway in both variants we have another problem – we change DI flow. We usually pass all dependencies into `Node` ready to use, but here we have to instantiate `ViewModel` inside `Node`, comparing to everything else that is done outside. Not like a bad thing, something that I want to come to later.

yes, that's what I was gonna do as a second option

mattinger:
The standard lifecycle viewModel functions for compose delegate to the composition local LocalViewModelStoreOwner.   Wouldn't it just be a matter of coming up with a proper implementation of this which was aware of the navigation graph, and setting up the composition local for each node?  (Yes, i know i'm making the implementation seem simpler than it probably would be)

I honestly wonder how compose navigation achieves this.

Update:  I took a look and it seems like NavBackStackEntry itself implements several interfaces including the LifecycleOwner and ViewModelStoreOwner.  So what i assume happens is that as each backstack entry is created, a completely new store is instantiated and registered as the local store owner.  And when it's popped off, it is completely cleared.  This makes a ton of sense to me obviously, and I think it could honestly be done in this library as well.

KovalevAndrey:
> The standard lifecycle viewModel functions for compose delegate to the composition local LocalViewModelStoreOwner. Wouldn't it just be a matter of coming up with a proper implementation of this which was aware of the navigation graph, and setting up the composition local for each node? (Yes, i know i'm making the implementation seem simpler than it probably would be)
> 
> I honestly wonder how compose navigation achieves this.
> 
> Update: I took a look and it seems like NavBackStackEntry itself implements several interfaces including the LifecycleOwner and ViewModelStoreOwner. So what i assume happens is that as each backstack entry is created, a completely new store is instantiated and registered as the local store owner. And when it's popped off, it is completely cleared. This makes a ton of sense to me obviously, and I think it could honestly be done in this library as well.

Hi @mattinger we're considering the same approach in this [draft pr](https://github.com/bumble-tech/appyx/pull/322/files) but the final decision whether to make it as a part of framework hasn't been made yet. 

We recently introduced [RetainedInstanceStore](https://github.com/bumble-tech/appyx/pull/361) which allows you to implement ViewModel support the way you described right now

PR #307, opened 2022-12-20
Bottom tab navigation sample

### LINKED ISSUE #305: Add bottom navigation demo to the sample app
We need to have a demo of how to implement a bottom tab nested navigation because it's pretty common pattern among apps.
https://m3.material.io/components/navigation-bar/guidelines

antonshilov:
## Description

- Changed the Spotlight demo ui to use material tab bar
- Added `SpotlightExt.current` method 

Fixes #305 

https://user-images.githubusercontent.com/8353068/208695717-33ff48df-737d-4156-b520-4ec893982379.mp4

## Check list

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 - [#287](https://github.com/bumble-tech/appyx/pull/287) – **Added**: `ImmutableList` has been added to avoid non-skippable compositions. 
 - [#289](https://github.com/bumble-tech/appyx/issues/289) – **Added**: Introduced `interop-rx3` for RxJava 3 support. This has identical functionality to `interop-rx2`.
 - [#298](https://github.com/bumble-tech/appyx/pulls/298) – **Updated**: ChildView documentation. `TransitionDescriptor` generics has been renamed to `NavTarget` and `State`
-- [#307](https://github.com/bumble-tech/appyx/pull/307) - **Added**: `Spotlight,current()` method to observe currently active `NavTarget.
+- [#307](https://github.com/bumble-tech/appyx/pull/307) - **Added**: `Spotlight.current()` method to observe currently active `NavTarget.
 
 ---
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - [#287](https://github.com/bumble-tech/appyx/pull/287) – **Added**: `ImmutableList` has been added to avoid non-skippable compositions. 
 - [#289](https://github.com/bumble-tech/appyx/issues/289) – **Added**: Introduced `interop-rx3` for RxJava 3 support. This has identical functionality to `interop-rx2`.
 - [#298](https://github.com/bumble-tech/appyx/pulls/298) – **Updated**: ChildView documentation. `TransitionDescriptor` generics has been renamed to `NavTarget` and `State`
+- [#307](https://github.com/bumble-tech/appyx/pull/307) - **Added**: `Spotlight,current()` method to observe currently active `NavTarget.
 
 ---
 
--- a/samples/sandbox/src/main/kotlin/com/bumble/appyx/sandbox/client/spotlight/SpotlightExampleNode.kt
+++ b/samples/sandbox/src/main/kotlin/com/bumble/appyx/sandbox/client/spotlight/SpotlightExampleNode.kt
@@ -84,7 +84,7 @@ class SpotlightExampleNode(
             screenState.value = Loading
             lifecycle.coroutineScope.launch {
                 delay(1000)
-                spotlight.updateElements(items = Item.getItemList())
+                spotlight.updateElements(items = Tab.getTabList())
                 screenState.value = Loaded
             }
         } else {
@@ -105,13 +105,13 @@ class SpotlightExampleNode(
     }
 
     @Parcelize
-    private enum class Item(val navTarget: NavTarget) : Parcelable {
+    private enum class Tab(val navTarget: NavTarget) : Parcelable {
         C1(Child1),
         C2(Child2),
         C3(Child3);
 
         companion object {
-            fun getItemList() = values().map { it.navTarget }
+            fun getTabList() = values().map { it.navTarget }
         }
     }
 
@@ -149,7 +149,7 @@ class SpotlightExampleNode(
         Scaffold(
             modifier = modifier.fillMaxSize(),
             floatingActionButtonPosition = FabPosition.Center,
-            floatingActionButton = { PageButtons(hasPrevious, hasNext) },
+            floatingActionButton = { PageButtons(hasPrevious.value, hasNext.value) },
             bottomBar = {
                 BottomTabs(currentTab)
             }
@@ -166,8 +166,8 @@ class SpotlightExampleNode(
     @Composable
     private fun BottomTabs(currentTab: State<NavTarget?>) {
         NavigationBar {
-            Item.values().forEach {
-                val selected = currentTab.value == it.navTarget
+            Tab.values().forEach { tab ->
+                val selected = currentTab.value == tab.navTarget
                 NavigationBarItem(
                     icon = {
                         Icon(
@@ -175,12 +175,12 @@ class SpotlightExampleNode(
                                 Icons.Filled.Favorite
                             else
                                 Icons.Outlined.FavoriteBorder,
-                            contentDescription = it.toString()
+                            contentDescription = tab.toString()
                         )
                     },
-                    label = { Text(it.toString()) },
+                    label = { Text(tab.toString()) },
                     selected = selected,
-                    onClick = { spotlight.activate(it) }
+                    onClick = { spotlight.activate(tab) }
                 )
             }
         }
@@ -188,8 +188,8 @@ class SpotlightExampleNode(
 
     @Composable
     private fun PageButtons(
-        hasPrevious: State<Boolean>,
-        hasNext: State<Boolean>
+        hasPrevious: Boolean,
+        hasNext: Boolean
     ) {
         Row(
             modifier = Modifier
@@ -200,11 +200,11 @@ class SpotlightExampleNode(
         ) {
             FilledIconButton(
                 onClick = { spotlight.previous() },
-                modifier = if (hasPrevious.value) Modifier.shadow(
+                modifier = if (hasPrevious) Modifier.shadow(
                     4.dp,
                     IconButtonDefaults.filledShape
                 ) else Modifier,
-                enabled = hasPrevious.value,
+                enabled = hasPrevious,
             ) {
                 Icon(
                     Icons.Filled.ArrowBack,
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - [#287](https://github.com/bumble-tech/appyx/pull/287) – **Added**: `ImmutableList` has been added to avoid non-skippable compositions. 
 - [#289](https://github.com/bumble-tech/appyx/issues/289) – **Added**: Introduced `interop-rx3` for RxJava 3 support. This has identical functionality to `interop-rx2`.
 - [#298](https://github.com/bumble-tech/appyx/pulls/298) – **Updated**: ChildView documentation. `TransitionDescriptor` generics has been renamed to `NavTarget` and `State`
+- [#307](https://github.com/bumble-tech/appyx/pull/307) - **Added**: `Spotlight,current()` method to observe currently active `NavTarget.
```
LachlanMcKee: `Spotlight.current()`?

samples/sandbox/src/main/kotlin/com/bumble/appyx/sandbox/client/spotlight/SpotlightExampleNode.kt
```diff
@@ -58,19 +71,19 @@ class SpotlightExampleNode(
     navModel = spotlight
 ) {
 
-    private val screenState = mutableStateOf<State?>(null)
+    private val screenState = mutableStateOf<ScreenState?>(null)
 
-    sealed class State {
-        object Loading : State()
-        object Loaded : State()
+    sealed class ScreenState {
```
LachlanMcKee, line 76: Why the rename?

samples/sandbox/src/main/kotlin/com/bumble/appyx/sandbox/client/spotlight/SpotlightExampleNode.kt
```diff
@@ -25,27 +38,27 @@ import com.bumble.appyx.core.node.Node
...
-import com.bumble.appyx.sandbox.client.spotlight.SpotlightExampleNode.State.Loaded
-import com.bumble.appyx.sandbox.client.spotlight.SpotlightExampleNode.State.Loading
+import com.bumble.appyx.sandbox.client.spotlight.SpotlightExampleNode.ScreenState.Loaded
+import com.bumble.appyx.sandbox.client.spotlight.SpotlightExampleNode.ScreenState.Loading
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.launch
 import kotlinx.parcelize.Parcelize
 
+/**
+ * Shows how to use spotlight to create a UI with BottomTabBar
```
LachlanMcKee, line 60: Does it make sense that this is a BottomTabBar sample? Why not put this in the sample app instead of sandbox?

samples/sandbox/src/main/kotlin/com/bumble/appyx/sandbox/client/spotlight/SpotlightExampleNode.kt
```diff
@@ -126,78 +139,91 @@ class SpotlightExampleNode(
...
                 navModel = spotlight
             )
+        }
+    }
 
+    @Composable
+    private fun BottomTabs(currentTab: State<NavTarget?>) {
+        NavigationBar {
+            Item.values().forEach {
+                val selected = currentTab.value == it.navTarget
```
LachlanMcKee: Give `it` a variable name? The `it`s are used quite a few times and might cause conflicts with other lambas.

samples/sandbox/src/main/kotlin/com/bumble/appyx/sandbox/client/spotlight/SpotlightExampleNode.kt
```diff
@@ -126,78 +139,91 @@ class SpotlightExampleNode(
...
+            }
         }
     }
 
     @Composable
-    private fun TextButton(text: String, enabled: Boolean = true, onClick: () -> Unit) {
-        Button(onClick = onClick, enabled = enabled, modifier = Modifier.padding(4.dp)) {
-            Text(text = text)
+    private fun PageButtons(
+        hasPrevious: State<Boolean>,
```
LachlanMcKee: Can you pass the booleans in rather than the state?

samples/sandbox/src/main/kotlin/com/bumble/appyx/sandbox/client/spotlight/SpotlightFaderThrough.kt
```diff
@@ -0,0 +1,89 @@
...
+    override fun createModifier(
+        modifier: Modifier,
+        transition: Transition<Spotlight.State>,
+        descriptor: TransitionDescriptor<T, Spotlight.State>
+    ): Modifier = modifier.composed {
+        val alpha = transition.animateFloat(
+            transitionSpec = {
+                when (targetState) {
+                    Spotlight.State.ACTIVE -> tween(
+                        durationMillis = 210,
```
LachlanMcKee: This is a rather random number. Maybe a constant would make sense? I assume it's a combination of 200ms plus 10ms from the `delayMillis`?

samples/sandbox/src/main/kotlin/com/bumble/appyx/sandbox/client/spotlight/SpotlightFaderThrough.kt
```diff
@@ -0,0 +1,89 @@
...
+                        delayMillis = 10,
+                        easing = FastOutSlowInEasing
+                    )
+
+                    Spotlight.State.INACTIVE_BEFORE,
+                    Spotlight.State.INACTIVE_AFTER -> tween(
+                        durationMillis = 90,
```
LachlanMcKee: 90ms is also strange. Is it also related to the 10ms? from above?

samples/sandbox/src/main/kotlin/com/bumble/appyx/sandbox/client/spotlight/SpotlightFaderThrough.kt
```diff
@@ -0,0 +1,89 @@
...
+                    Spotlight.State.ACTIVE -> 1f
+                    else -> 0f
+                }
+            }, label = ""
+        )
+        val scale = transition.animateFloat(
+            transitionSpec = {
+                when (targetState) {
+                    Spotlight.State.ACTIVE -> tween(
+                        durationMillis = 210,
```
LachlanMcKee: These values seem to be reused. A constant would be a good idea!

samples/sandbox/src/main/kotlin/com/bumble/appyx/sandbox/client/spotlight/SpotlightFaderThrough.kt
```diff
@@ -0,0 +1,89 @@
...
+            alpha(alpha.value)
+        }
+
+    }
+}
+
+@Composable
+fun <T> rememberSpotlightFaderThrough(): ModifierTransitionHandler<T, Spotlight.State> = remember {
+    SpotlightFaderThrough()
+}
```
LachlanMcKee: This will likely be causing a detekt issue.

samples/sandbox/src/main/kotlin/com/bumble/appyx/sandbox/client/spotlight/SpotlightExampleNode.kt
antonshilov: State was conflicting with compose.runtime.State

samples/sandbox/src/main/kotlin/com/bumble/appyx/sandbox/client/spotlight/SpotlightExampleNode.kt
antonshilov: Hmm, not sure to be honest. This demo was initially placed in sandbox so I decided to leave it there for now

samples/sandbox/src/main/kotlin/com/bumble/appyx/sandbox/client/spotlight/SpotlightFaderThrough.kt
antonshilov: These timings are taken from material spec for this kind of transition. Overall duration should be 300ms, more details here https://m2.material.io/design/motion/the-motion-system.html#fade-through
image

antonshilov: 10 ms delay was a mistake though, thanks for highlighting

samples/sandbox/src/main/kotlin/com/bumble/appyx/sandbox/client/spotlight/SpotlightExampleNode.kt
LachlanMcKee: I believe that's a problem that we have in a lot of places in the codebase. We should probably set a standard rather than having this sample be different all other nodes.

LachlanMcKee (approved):

zsoltk:
@antonshilov Thanks for the PR! 

I'm coming in late since this is already merged, a few considerations for the future:

1. Baked-in time values don't seem like a good idea, they should be configurable by client code.
2. Baked-in tween as a spec doesn't feel right either – it should probably be spring so that it's interruptable + would be nice if client code could override it in this case too.
3. Rather than a property-by-property approach in the transition handler I've found it to be a nice pattern to use a Props class to group related properties together. It makes the code much more digestible + maintainable. I think I used this approach in Cards & Promoter (can't check now).

All of this matters as people will most probably copy-paste this class.

Agreed with @LachlanMcKee that we should also consider examples in the `:app` rather than `:sandbox` (but let's treat it as a well-organised, nice and high quality front facing app rather than a dump that sandbox is).

jeddchoi:
Hi, i'm investigating this library for bottom navigation use. 
I want to make bottom navigation like [Google's example app_Now in Android](https://github.com/android/nowinandroid/tree/main).


1. I'm not sure Spotlight is similar to above example app.
For NowInAndroid, there are 3 tabs-bottom navigation and when bottom navigation item is selected, code below is executed.

```kotlin
navController.navigate("some-route") {
    // Pop up to the start destination of the graph to
    // avoid building up a large stack of destinations
    // on the back stack as users select items
    popUpTo(
        navController.graph.findStartDestination().id
    ) {
        saveState = true
    }
    // Avoid multiple copies of the same destination when
    // reselecting the same item
    launchSingleTop = true
    // Restore state when reselecting a previously selected item
    restoreState = true
}
```


2. I know that document said 
> "Implements a mechanism analogous to a view pager; has a single active element ("it's in the spotlight", hence the name), but unlike the back stack, it does not remove other elements.".  

I'm afraid that it would take up a lot of memory when each tab's screen has many contents with Spotlight.
Also, I wonder that Spotlight would consider all lifecycle-things properly such as kotlin flow, coroutines in each screen's ViewModel(per tab). Would it pause and collect the flow properly?


3. Plus, `Spotlight.previous()` or `next()` is not desired usually for bottom navigation, I think.


Am I right in thinking? Should I make my own navModel class with BackStack? 
I saw [this project(which you mentioned in doc)](https://github.com/jbreitfeller-sfix/appyx-playground) also made bottom navigation with BackStack.

Because I don't have much experience in Android dev, I apologize in advance.

PR #311, opened 2023-01-02
[#310] Ensure NavTarget and State are parcelable

### LINKED ISSUE #310: NavKey navTarget parcelable is not enforced
At the moment the NavKey `navTarget` argument is not enforced to be parcelable. It uses `@RawValue` to avoid the compiler complaining.

Unfortunately this means it is possible to introduce code that will crash at runtime (there is currently an issue in the newly introduced `IndexedBackStack` causing a crash when the app is backgrounded.

By adding more constraints to the generics we can ensure that we have compile time safety.

LachlanMcKee:
## Description
Fixes #310

## Check list

### DIFF
--- a/samples/app/src/main/kotlin/com/bumble/appyx/app/node/backstack/app/indexedbackstack/IndexedBackStack.kt
+++ b/samples/app/src/main/kotlin/com/bumble/appyx/app/node/backstack/app/indexedbackstack/IndexedBackStack.kt
@@ -1,12 +1,14 @@
 package com.bumble.appyx.app.node.backstack.app.indexedbackstack
 
+import android.os.Parcelable
 import com.bumble.appyx.core.navigation.BaseNavModel
 import com.bumble.appyx.core.navigation.NavElements
 import com.bumble.appyx.core.navigation.NavKey
 import com.bumble.appyx.core.navigation.Operation
 import com.bumble.appyx.core.state.SavedStateMap
+import kotlinx.parcelize.Parcelize
 
-class IndexedBackStack<NavTarget : Any>(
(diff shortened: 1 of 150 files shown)


samples/app/src/main/kotlin/com/bumble/appyx/app/node/backstack/app/indexedbackstack/IndexedBackStack.kt
```diff
@@ -15,14 +17,20 @@ class IndexedBackStack<NavTarget : Any>(
     savedStateMap = savedState
 ) {
 
-    sealed interface State {
+    sealed interface State : Parcelable {
```
LachlanMcKee, line 20: This was a bug introduced due to a lack of compile-time contracts

LachlanMcKee:
This is definitely a breaking change. We can likely fix the problems (samples that do not use Parcelable) in a different PR, but I think it's worthwhile adding this to avoid issues in the future

LachlanMcKee:
If we don't want to go down this path (I don't see why we shouldn't, other than the fact that we have to make the enums parcelable), we MAY be able to solve this with a `detekt` rule (though that would mean that the library may not work as advertised if developers don't use `detekt`)

CherryPerry:
The original intention was to allow different types like primitives and serialisables.

We can potentially introduce runtime checks `class.java.isPrimitive || class.java.implementsParcelableOrSerializable` as an alternative which we can perform in the constructor of `NavElement`.

Not against your approach.

LachlanMcKee:
@CherryPerry that could be a good alternative as well. Let's discuss tomorrow!

PR #317, opened 2023-01-13
Clean up gradle catalogue and use compose BOM

KovalevAndrey:
## Description

Simplification compose version management by using [BOM](https://developer.android.com/jetpack/compose/setup#bom-version-mapping)

## Check list

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -33,6 +33,15 @@ androidx-test-junit = "androidx.test.ext:junit:1.1.3"
 androidx-test-rules = "androidx.test:rules:1.4.0"
 coil-compose = { module = "io.coil-kt:coil-compose", version.ref = "coil" }
 compose-bom = { group = "androidx.compose", name = "compose-bom", version.ref = "composeBom" }
+# compose versions are resolved by BOM
+compose-animation-core = { module = "androidx.compose.animation:animation-core" }
+compose-foundation-layout = { module = "androidx.compose.foundation:foundation-layout" }
+compose-foundation = { module = "androidx.compose.foundation:foundation" }
+compose-material = { module = "androidx.compose.material3:material3" }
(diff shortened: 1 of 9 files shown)


libraries/core/build.gradle.kts
```diff
@@ -35,23 +35,27 @@ dependencies {
     api(project(":libraries:customisations"))
     api(libs.kotlin.coroutines.android)
     api(libs.androidx.lifecycle.common)
-    api(libs.compose.animation.core)
-    api(libs.compose.runtime)
+    api("androidx.compose.animation:animation-core")
```
LachlanMcKee: Is this the right way to approach this? I'm not a fan of having two different styles here. Does version catalogs not work with BOMs?

You lose the type safety this way

LachlanMcKee: I would suggest looking at this solution: https://stackoverflow.com/questions/67192785/gradle-7-0-version-catalog-for-maven-bom

gradle/libs.versions.toml
```diff
@@ -8,7 +8,7 @@ accompanist = "0.28.0"
 androidx-lifecycle = "2.5.1"
 androidx-navigation-compose = "2.5.1"
 coil = "2.2.1"
-compose = "1.3.0"
+composeBom = "2023.01.00"
 composeCompiler = "1.3.1"
```
LachlanMcKee, line 12: Is the compiler version in the BOM somehow?

KovalevAndrey: https://developer.android.com/jetpack/compose/setup#why_isnt_the_compose_compiler_library_included_in_the_bom

libraries/core/build.gradle.kts
KovalevAndrey: I put them back into catalogue

LachlanMcKee (approved):

PR #319, opened 2023-01-16
Add interactions module

zsoltk:


appyx-interactions/build.gradle.kts
```diff
@@ -0,0 +1,47 @@
...
+
+        testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
+    }
+    buildFeatures {
+        compose = true
+    }
+    composeOptions {
+        kotlinCompilerExtensionVersion = libs.versions.composeCompiler.get()
+    }
+    testOptions {
```
CherryPerry: No actual need for this block.

appyx-interactions/.gitignore
```diff
@@ -0,0 +1 @@
+/build
```
CherryPerry: We do not have .gitignore files in other modules.

appyx-interactions/proguard-rules.pro
```diff
@@ -0,0 +1,21 @@
+# Add project specific ProGuard rules here.
```
CherryPerry: Can be also removed.

settings.gradle.kts
```diff
@@ -17,6 +17,7 @@ dependencyResolutionManagement {
 enableFeaturePreview("VERSION_CATALOGS")
 
 include(
+    ":appyx-interactions",
```
LachlanMcKee, line 20: Why is this outside of the libraries/samples folders?

appyx-interactions/build.gradle.kts
```diff
@@ -0,0 +1,47 @@
...
+            // interface method default implementation
+            it.exclude("**/*\$DefaultImpls.class")
+        }
+    }
+}
+
+dependencies {
+    val composeBom = platform(libs.compose.bom)
+
+    api(composeBom)
```
LachlanMcKee: You're currently not using these, the build will fail (as we ensure that all dependencies are used as part of CI). You should remove all the dependencies for now.

LachlanMcKee (dismissed):

appyx-interactions/build.gradle.kts
```diff
@@ -0,0 +1,47 @@
+plugins {
+    id("com.android.library")
+    id("kotlin-android")
```
LachlanMcKee, line 3: Is this a library or a sample? is it going to be published?

settings.gradle.kts
zsoltk: I imagine splitting to two top-level modules, appyx-interactions and appyx-navigation. Then you can have libraries/samples inside of each. I haven't gone the full way yet with appyx-navigation to unblock us working on the interactions.

appyx-interactions/build.gradle.kts
zsoltk: Library. Eventually published, not for a while yet.

settings.gradle.kts
LachlanMcKee: there is no such thing as a top level module. Do you mean this to be a folder? if so, please don't make this a gradle module.

zsoltk: For now it will be used as a draft place to unblock us. We can always rearrange as we go. No harm is done, this isn't affecting 1.x or publication.

zsoltk:
Closing in favor of #320 instead.

PR #325, opened 2023-01-21
Ensure node was built before using PermanentChild API

KovalevAndrey:
## Description

Using PermanentChild in UI testing leads to a crash due to `DummyParentNode` not being built.

## Check list

### DIFF
--- a/libraries/core/src/main/kotlin/com/bumble/appyx/core/node/Node.kt
+++ b/libraries/core/src/main/kotlin/com/bumble/appyx/core/node/Node.kt
@@ -75,7 +75,7 @@ open class Node(
             field = value
         }
 
-    protected var wasBuilt = false
+    private var wasBuilt = false
 
     val id = getNodeId(buildContext)
 
@@ -229,10 +229,6 @@ open class Node(
 
     private fun handleUpNavigationByPlugins(): Boolean =
         plugins<UpNavigationHandler>().any { it.handleUpNavigation() }
-
-    protected fun ensureNodeWasBuilt() {
-        require(wasBuilt) { "Node hasn't been built yet" }
-    }
 
     companion object {
         private const val NODE_ID_KEY = "node.id"
--- a/libraries/core/src/main/kotlin/com/bumble/appyx/core/node/Node.kt
+++ b/libraries/core/src/main/kotlin/com/bumble/appyx/core/node/Node.kt
@@ -75,7 +75,7 @@ open class Node(
             field = value
         }
 
-    private var wasBuilt = false
+    protected var wasBuilt = false
 
     val id = getNodeId(buildContext)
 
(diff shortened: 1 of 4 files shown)


libraries/core/src/main/kotlin/com/bumble/appyx/core/node/Node.kt
```diff
@@ -75,7 +75,7 @@ open class Node(
             field = value
         }
 
-    private var wasBuilt = false
+    protected var wasBuilt = false
```
CherryPerry: I do not think it should be part of public API.

libraries/core/src/main/kotlin/com/bumble/appyx/core/node/Node.kt
```diff
@@ -230,6 +230,10 @@ open class Node(
     private fun handleUpNavigationByPlugins(): Boolean =
         plugins<UpNavigationHandler>().any { it.handleUpNavigation() }
 
+    protected fun ensureNodeWasBuilt() {
```
CherryPerry: We already have this check inside lifecycle listener of the node, it does not work?

libraries/core/src/main/kotlin/com/bumble/appyx/core/node/ParentNode.kt
```diff
@@ -101,6 +101,7 @@ abstract class ParentNode<NavTarget : Any>(
         navTarget: NavTarget,
         decorator: @Composable (child: ChildRenderer) -> Unit
     ) {
+        ensureNodeWasBuilt()
```
CherryPerry: Looks like a crutch for this particular case, makes sense to invoke everywhere.

libraries/testing-ui/src/main/kotlin/com/bumble/appyx/testing/ui/rules/AppyxViewTestRule.kt
```diff
@@ -47,7 +48,7 @@ open class AppyxViewTestRule<View : NodeView>(
     override fun beforeActivityLaunched() {
         AppyxTestActivity.composableView = {
             CompositionLocalProvider(
-                LocalNode provides DummyParentNode<Any>(),
+                LocalNode provides DummyParentNode<Any>().build(),
```
CherryPerry, line 51: I am fine keeping this as the only change of PR.

libraries/core/src/main/kotlin/com/bumble/appyx/core/node/Node.kt
KovalevAndrey: nope, we are not using NodeHost for view testing which propagates lifecycle events. Lifecycle never makes it to `onCreate` to enable that check

libraries/testing-ui/src/main/kotlin/com/bumble/appyx/testing/ui/rules/AppyxViewTestRule.kt
KovalevAndrey: ok. let's keep only this change otherwise it gets ugly to make this check only for view testing use case

libraries/core/src/main/kotlin/com/bumble/appyx/core/node/Node.kt
KovalevAndrey: reverted

CherryPerry (approved):

PR #330, opened 2023-01-26
Proper lifecycle support

KovalevAndrey:
1. ParentNode works with InteractionModel instead of TransitionModel
2. On/off screen support
3. Lifecycle support
4. Proper children creation considering Keep/suspend mode
5. Cards mechanism reworked

## Description

...

## Check list


appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/transitionmodel/cards/interpolator/CardsProps.kt
```diff
@@ -40,37 +40,44 @@ class CardsProps<NavTarget : Any>(
...
         rotationZ = -45f,
         zIndex = 2f,
+        isVisible = false
     )
 
     private val voteLike = top.copy(
         positionalOffsetX = (voteCardPositionMultiplier * width).dp,
         rotationZ = 45f,
         zIndex = 2f,
+        isVisible = true
```
KovalevAndrey: false

PR #336, opened 2023-01-31
Relax ChildAware API requirements

CherryPerry:
## Description

ChildAware API should allow any type instead of `Node` subtypes.
The current implementation is too permissive and does not allow to use interfaces as internode communication contract:

```kotlin
interface Child : Connectable<Input, Output> {
  sealed class Input
  sealed class Output
}

// We can't make ChildNode internal like we did in RIBs
class ChildNode(...): Node(...), Child by NodeConnector() {} 
```

```kotlin
// Somewhere in the parent node
whenChildrenAttached<Child>(...) // <-- Can't use Child interface here right now
whenChildrenAttached<ChildNode>(...) // <-- Have to use node instead with much wider public API
```

With this change it is possible to use `Child` interface to setup node connections, but **still impossible** to make `ChildNode` class internal. To do it we need to change `Builder` classes too, either introduce base `Node` interface like `Rib` interface or come up with something else.

## Check list

### DIFF
--- a/libraries/core/src/main/kotlin/com/bumble/appyx/core/children/ChildAwareImpl.kt
+++ b/libraries/core/src/main/kotlin/com/bumble/appyx/core/children/ChildAwareImpl.kt
@@ -53,7 +53,7 @@ class ChildAwareImpl<N : Node> : ChildAware<N> {
             }
     }
 
-    override fun <T : Node> whenChildAttached(
+    override fun <T : Any> whenChildAttached(
         child: KClass<T>,
         callback: ChildCallback<T>
     ) {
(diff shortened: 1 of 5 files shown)


libraries/core/src/main/kotlin/com/bumble/appyx/core/children/ChildAwareImpl.kt
```diff
@@ -53,7 +53,7 @@ class ChildAwareImpl<N : Node> : ChildAware<N> {
             }
     }
 
-    override fun <T : Node> whenChildAttached(
+    override fun <T : Any> whenChildAttached(
         child: KClass<T>,
         callback: ChildCallback<T>
     ) {
```
KovalevAndrey, line 59: Maybe check if `child` is a `Node` to fail early? Rather that never get a callback?

CherryPerry: But it is not, that the main point. In my sample `Child` is just an interface, like we have in our templates.

KovalevAndrey (approved):
I don't mind, but we'll never know if we accidentally pass wrong class

PR #342, opened 2023-02-07
Dynamic props visibility calculation

KovalevAndrey:
## Description

1. Removing destroyed elements for every next segment
2. Dynamically calculating props visibility

## Check list


appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/transitionmodel/backstack/interpolator/BackStackSlider.kt
```diff
@@ -21,19 +22,22 @@ import kotlinx.coroutines.async
...
     data class Props(
         val offset: Offset = Offset(DpOffset(0.dp, 0.dp)),
         val alpha: Alpha = Alpha(value = 1f),
         val offsetMultiplier: Int = 1,
-        override val isVisible: Boolean = true
+        val screenWidth: Dp
     ) : Interpolatable<Props>, HasModifier, BaseProps, Animatable<Props> {
 
+        override val isVisible: Boolean
+            get() = alpha.value >= 0.0f && offset.value.x < screenWidth && offset.value.x > -screenWidth
```
zsoltk: Not entirely true unless we also take the element's own width into account (or refactor to an `activeIndex`-based approach here too). I'm fine with this now, but let's at least leave a comment.

appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/transitionmodel/cards/interpolator/CardsProps.kt
```diff
@@ -53,9 +53,13 @@ class CardsProps<NavTarget : Any>(
         ),
         val rotationZ: RotationZ = RotationZ(value = 0f),
         val zIndex: ZIndex = ZIndex(value = 0f),
-        override val isVisible: Boolean = false
+        private val width: Float,
     ) : Interpolatable<Props>, HasModifier, Animatable<Props>, BaseProps {
 
+        override val isVisible: Boolean
+            get() = scale.value >= 0.0f && positionalOffsetX.value.x > (-voteCardPositionMultiplier * width).dp &&
```
zsoltk, line 60: Also need to take `zIndex` into account, as the bottom card will be covered by the top card. Wouldn't it be easier to rely (fully or partially) on an enum value in this class instead of trying to incorporate all UI properties? It's a simplification that in most cases would still be right.

appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/transitionmodel/spotlight/SpotlightModel.kt
```diff
@@ -48,13 +49,17 @@ class SpotlightModel<NavTarget : Any>(
...
+            positions = items.map {
+                State.Position(
+                    elements = mapOf(it.asElement() to STANDARD)
+                )
+            },
             activeIndex = initialActiveIndex,
             activeWindow = initialActiveWindow
         )
 
+    override fun State<NavTarget>.removeDestroyedElements(): State<NavTarget> = this
```
zsoltk, line 62: TODO? This model supports destroyed element status, so we should clear those.

appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/transitionmodel/cards/interpolator/CardsProps.kt
KovalevAndrey: > Also need to take `zIndex` into account, as the bottom card will be covered by the top card. Wouldn't it be easier to rely (fully or partially) on an enum value in this class instead of trying to incorporate all UI properties? It's a simplification that in most cases would still be right.

I think it's fine if we draw 2 cards instead of one when top card is not dragged

appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/transitionmodel/spotlight/interpolator/SpotlightSlider.kt
```diff
@@ -62,9 +64,24 @@ class SpotlightSlider<NavTarget : Any>(
...
+        val scrollValue: () -> Float,
+        private val width: Dp,
+        private val activeWindow: Float
     ) : Interpolatable<Props>, HasModifier, Animatable<Props>, BaseProps {
 
+        private val activeWindowOffset = (activeWindow * width.value).dp
+
+        override val isVisible: Boolean
+            get() {
+                val currentOffset = (scrollValue() * width.value).dp
```
zsoltk, line 78: Please check against my changes, the scroll offset should already be added as a displacement, so you don't need to add it a second time.

samples/appyx-navigation/src/main/kotlin/com/bumble/appyx/appyxnavigation/node/spotlight/SpotlightNode.kt
```diff
@@ -50,7 +50,7 @@ class SpotlightNode(
             initialActiveIndex = 0f,
             initialActiveWindow = 1f
         ),
-        interpolator = { SpotlightSlider(it, coroutineScope) }
+        interpolator = { SpotlightSlider(it, coroutineScope, activeWindow = 0f) }
```
zsoltk: 0?

KovalevAndrey: typo, 1f

PR #343, opened 2023-02-08
Geometry fixes

zsoltk:
Cleans up a lot of custom code by moving geometry handling to the abstract base class, based on a mapping.
Fixes edge cases of how geometry should react to update vs keyframe.


appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/transitionmodel/BaseInterpolator.kt
```diff
@@ -115,6 +112,14 @@ abstract class BaseInterpolator<NavTarget : Any, ModelState, Props>(
         val fromProps = fromState.toProps()
         val targetProps = targetState.toProps()
 
+        runBlocking {
```
KovalevAndrey: to avoid ANRs, let's launch it in the scope provided to `BaseInteroplator`

PR #348, opened 2023-02-09
Screenshot tests suite

CherryPerry:
## Description

Plugin to run screenshot tests.

How to use:
1. Apply `id("appyx-screenshots")` in `build.gradle.kts`
2. Use the following structure for tests:
```kotlin
@get:Rule
val nameRule = TestName()

@Test
fun test() {
   bitmap.writeToTestStorage("${javaClass.simpleName}_${nameRule.methodName}")
}
```
3. Run either `screenshotTestsCompareBaseline` or `screenshotTestsRecordBaseline` task.

Requirements: `nodejs`, `npm`, `npx`

## Check list

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -75,7 +75,7 @@ jobs:
   instrumentation-tests:
     name: Instrumentation tests
     runs-on: macOS-latest
-    timeout-minutes: 30
+    timeout-minutes: 60
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-java@v3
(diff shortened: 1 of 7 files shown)


LachlanMcKee:
Why not use paparazzi? This seems like it will be quite a lot slower unfortunately

LachlanMcKee:
I had attempted to do this here: https://github.com/bumble-tech/appyx/pull/244

CherryPerry:
@vladcipariu91 asked for real screenshots because paparazzi did not work for his case.

LachlanMcKee:
It would be good to discuss what the problem is. perhaps it's the same problem I faced?

vladcipariu91:
as discussed with @LachlanMcKee we can merge this and look into a fix for Paparazzi later

PR #356, opened 2023-02-14
Threshold per field

zsoltk:
Allows defining sensible thresholds for spring animations per property. As a result, animations settle much, much quicker instead of oscillating forever on microscopic differences.


KovalevAndrey:
> Allows defining sensible thresholds for spring animations per property. As a result, animations settle much, much quicker instead of oscillating forever on microscopic differences.

Shouldn't we just supply proper animation spec with no bouncing instead?

KovalevAndrey (approved):

PR #359, opened 2023-02-15
Back navigation

antonshilov:
## Description

basic implementation of backpress handling for BackStack. Does not include plugins

## Check list


appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/transitionmodel/backstack/BackStack.kt
```diff
@@ -25,4 +28,16 @@ class BackStack<NavTarget : Any>(
...
-)
+) {
+    override fun handleBackNavigation(): Boolean {
+        val pop = Pop<NavTarget>()
+        //todo find a better way to check if operation is applicable
+        return if (pop.isApplicable(model.output.value.currentTargetState)) {
+            operation(Pop())
+            true
+        } else false
+    }
```
zsoltk: We could apply 1.x approach here.

antonshilov: Do you mean backpress strategy?

zsoltk: Yes

PR #361, opened 2023-02-18
Introduced RetainedInstanceStore to keep data during config change

LachlanMcKee:
## Description
Copied the RetainedInstanceStore concept from RIBs so that objects can be retained during a configuration change.

Also re-added the MVICore screens to the sandbox to demonstrate the behaviour

## Check list

### DIFF
--- a/libraries/core/src/main/kotlin/com/bumble/appyx/core/modality/BuildContext.kt
+++ b/libraries/core/src/main/kotlin/com/bumble/appyx/core/modality/BuildContext.kt
@@ -1,9 +1,11 @@
 package com.bumble.appyx.core.modality
 
+import com.bumble.appyx.core.state.MutableSavedStateMap
+import com.bumble.appyx.core.state.SavedStateMap
 import com.bumble.appyx.utils.customisations.NodeCustomisation
 import com.bumble.appyx.utils.customisations.NodeCustomisationDirectory
 import com.bumble.appyx.utils.customisations.NodeCustomisationDirectoryImpl
-import com.bumble.appyx.core.state.SavedStateMap
+import java.util.UUID
 
 data class BuildContext(
     val ancestryInfo: AncestryInfo,
@@ -11,6 +13,8 @@ data class BuildContext(
     val customisations: NodeCustomisationDirectory,
 ) {
     companion object {
+        private const val IDENTIFIER_KEY = "build.context.identifier"
+
         fun root(
             savedStateMap: SavedStateMap?,
             customisations: NodeCustomisationDirectory = NodeCustomisationDirectoryImpl()
@@ -22,6 +26,18 @@ data class BuildContext(
             )
     }
 
-    fun <T : NodeCustomisation> getOrDefault(defaultCustomisation: T) : T =
+    val identifier: String by lazy {
+        if (savedStateMap == null) {
+            UUID.randomUUID().toString()
+        } else {
+            savedStateMap[IDENTIFIER_KEY] as String? ?: error("onSaveInstanceState() was not called")
+        }
+    }
+
+    fun <T : NodeCustomisation> getOrDefault(defaultCustomisation: T): T =
         customisations.getRecursivelyOrDefault(defaultCustomisation)
+
+    fun onSaveInstanceState(state: MutableSavedStateMap) {
+        state[IDENTIFIER_KEY] = identifier
+    }
 }
--- a/libraries/core/src/main/kotlin/com/bumble/appyx/core/store/RetainedInstanceStore.kt
+++ b/libraries/core/src/main/kotlin/com/bumble/appyx/core/store/RetainedInstanceStore.kt
@@ -1,0 +1,35 @@
+package com.bumble.appyx.core.store
+
+/**
+ * A simple storage to preserve any objects during configuration change events.
+ * `factory` function will be invoked immediately on the same thread
+ * only if an object of the same class within the same Node does not exist.
+ *
+ * The framework will manage the lifecycle of provided objects
+ * and invoke `disposer` function to destroy objects properly.
+ *
+ * Sample usage:
+ * ```kotlin
+ * val feature = RetainedInstanceStore.get(
+ *     storeId = buildContext.identifier,
+ *     factory = { FeatureImpl() },
+ *     disposer = { feature.dispose() }
+ * }
+ * ```
+ * or
+ *  * ```kotlin
+ *  * val feature = buildContext.getRetainedInstance(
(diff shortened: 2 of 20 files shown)


CherryPerry:
Originally I wanted to suggest to adopt https://github.com/bumble-tech/appyx/pull/322 instead, but after I checked API I do not want anymore (ViewModels API is a hell, you can't just pass ViewModel factory as a lambda).

But my suggestion is still to apply its approach regarding `ViewModelStoreOwner`. I do not like that and how we moved ID management from `Node` itself to `BuildParams`. So instead, what do you think if we make it in the same way as `ViewModelStoreOwner` – provide `RetainedInstanceStore` API only directly from `Node` and not outside. We can capture all required parameters in `Builder` and pass them as a lamda to `Node`. From the first look it worse, but it is a good opportunity to 1) to do less potential API breaking changes now and in the future, we can always to go from inside-node approach to outside-node approach (but not other way) 2) check if direct dependency injection makes sense (partially dropping `Builder` and completly `Interactor` class as it won't have access to the feature anymore).

Would like to invite more people to discuss.

LachlanMcKee:
@CherryPerry I'm happy if we can come up with another solution. This PR allows Devs to retained objects within the builder which is useful of you're using MVICore.

The ViewModel approach in the other PR doesn't seem to have this functionality.

Perhaps we could support both? As devs using the builder/interactor approach need a solution as well

libraries/core/src/main/kotlin/com/bumble/appyx/core/modality/BuildContext.kt
```diff
@@ -22,6 +26,18 @@ data class BuildContext(
             )
     }
 
-    fun <T : NodeCustomisation> getOrDefault(defaultCustomisation: T) : T =
+    val identifier: String by lazy {
```
LachlanMcKee, line 29: Make this internal?

LachlanMcKee: Add a comment at the very least saying to not use it, Incubating?

libraries/core/src/main/kotlin/com/bumble/appyx/core/store/RetainedInstanceStoreImpl.kt
```diff
@@ -0,0 +1,28 @@
+package com.bumble.appyx.core.store
+
+import kotlin.reflect.KClass
+
+internal class RetainedInstanceStoreImpl : RetainedInstanceStore {
+
+    private val map: MutableMap<String, MutableMap<KClass<*>, ValueHolder<*>>> = HashMap()
+
+    @Suppress("UNCHECKED_CAST")
+    override fun <T : Any> get(nodeId: String, clazz: KClass<*>, disposer: (T) -> Unit, factory: () -> T): T =
```
LachlanMcKee: Use a string key instead of the class so that multiple instance of the same class can be used.

libraries/core/src/main/kotlin/com/bumble/appyx/core/modality/BuildContext.kt
LachlanMcKee: We cannot make this internal unfortunately. Not sure if I should mark this with a comment as this will be part of the 1.0 API

CherryPerry (approved):

libraries/core/src/test/kotlin/com/bumble/appyx/core/store/RetainedInstanceStoreTest.kt
```diff
@@ -46,6 +46,44 @@ class RetainedInstanceStoreTest {
         assertEquals(2L, longValue)
     }
 
+    @Test
+    fun `GIVEN object stored with int type WHEN second object stored with int type THEN second factory not called`() {
```
KovalevAndrey: I don't think we should test it. We're not testing `RetainedInstanceStore` contract here but internal implementation. 
[This test](https://github.com/bumble-tech/appyx/pull/361/commits/5d629d2a2dfdc180c6363c920b9177762ca1151e#diff-40f24b8c2f8f8ef099fcc890959e97345f158cad3c9a27dfa8eb03d1f06140caR76) covers this particular scenario

libraries/core/src/main/kotlin/com/bumble/appyx/core/store/RetainedInstanceStore.kt
```diff
@@ -0,0 +1,35 @@
...
+ *  *     factory = { FeatureImpl() },
+ *  *     disposer = { feature.dispose() }
+ *  * }
+ *  * ```
+ */
+interface RetainedInstanceStore {
+
+    fun <T : Any> get(storeId: String, key: String, disposer: (T) -> Unit = {}, factory: () -> T): T
+
+    fun removeAll(storeId: String)
```
KovalevAndrey: [nit]
```suggestion
    fun clearStore/removeStore(storeId: String)
```

libraries/core/src/test/kotlin/com/bumble/appyx/core/store/RetainedInstanceStoreTest.kt
```diff
@@ -0,0 +1,110 @@
...
+    fun `GIVEN object stored WHEN get with same identifier THEN factory not called`() {
+        var factoryCalled = false
+        store.get(identifier) { Any() }
+
+        store.get(identifier) {
+            factoryCalled = true
+            Any()
+        }
+
+        assertFalse(factoryCalled)
```
KovalevAndrey, line 34: Same as below, we don't test the contract here

libraries/core/src/test/kotlin/com/bumble/appyx/core/store/RetainedInstanceStoreTest.kt
LachlanMcKee: fair enough. removed

libraries/core/src/test/kotlin/com/bumble/appyx/core/store/RetainedInstanceStoreTest.kt
LachlanMcKee: fair enough. removed

libraries/core/src/main/kotlin/com/bumble/appyx/core/store/RetainedInstanceStore.kt
LachlanMcKee: renamed to `clearStore`

KovalevAndrey (approved):

PR #362, opened 2023-02-20
Fix flickering bug by updating segment progress before main progress

antonshilov:
## Description

Demo of fixed bug

https://user-images.githubusercontent.com/8353068/220062123-d6f48692-7584-4c87-a7b4-71fa18debd63.mp4

### DIFF
--- a/appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/Keyframes.kt
+++ b/appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/Keyframes.kt
@@ -84,8 +84,8 @@ data class Keyframes<ModelState>(
         val currentProgress = this.progress.toInt()
         val progress = progress.coerceIn(0f, maxProgress)
         Logger.log("Keyframes", "$progress")
-        progressFlow.value = progress
         segmentProgress.value = progress - currentIndex
+        progressFlow.value = progress
         if (progress.toInt() > currentProgress) {
             Logger.log("Keyframes", "onTransitionFinished()")
             onTransitionFinished(currentSegment.fromState)
--- a/appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/Keyframes.kt
+++ b/appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/Keyframes.kt
@@ -5,6 +5,7 @@ import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.filter
+import kotlinx.coroutines.flow.filterNotNull
 import kotlinx.coroutines.flow.map
 
 data class Keyframes<ModelState>(
@@ -47,7 +48,7 @@ data class Keyframes<ModelState>(
     }
 
     fun getSegmentProgress(segmentIndex: Int): Flow<Float> =
-        progressFlow.map { it.toSegmentProgress(segmentIndex) }.filter { it in 0f..1f }
+        progressFlow.map { it.toSegmentProgress(segmentIndex) }.filterNotNull()
 
     val progress: Float
         get() = progressFlow.value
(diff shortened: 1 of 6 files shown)


appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/Keyframes.kt
```diff
@@ -84,8 +84,8 @@ data class Keyframes<ModelState>(
         val currentProgress = this.progress.toInt()
         val progress = progress.coerceIn(0f, maxProgress)
         Logger.log("Keyframes", "$progress")
-        progressFlow.value = progress
         segmentProgress.value = progress - currentIndex
+        progressFlow.value = progress
```
KovalevAndrey, line 91: let's leave a comment to prevent this from happening in the future?

KovalevAndrey (approved):

appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/Keyframes.kt
zsoltk: Not just comment, we should back this up by a test

appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/Keyframes.kt
```diff
@@ -44,7 +46,8 @@ data class Keyframes<ModelState>(
         queue[it].targetState
     }
 
-    val segmentProgress = MutableStateFlow(progressFlow.value - currentIndex)
+    fun getSegmentProgress(segmentIndex: Int): Flow<Float> =
+        progressFlow.map { it.toSegmentProgress(segmentIndex) }.filter { it in 0f..1f }
```
zsoltk: The `{ it in 0f..1f }` could (and probably should) be moved into `toSegmentProgress`, as it semantically already shouldn't go beyond `1f`.

appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/Keyframes.kt
```diff
@@ -85,11 +88,11 @@ data class Keyframes<ModelState>(
...
         Logger.log("Keyframes", "$progress")
         progressFlow.value = progress
-        segmentProgress.value = progress - currentIndex
         if (progress.toInt() > currentProgress) {
             Logger.log("Keyframes", "onTransitionFinished()")
             onTransitionFinished(currentSegment.fromState)
         }
     }
 }
+internal fun Float.toSegmentProgress(segmentIndex: Int) = this - segmentIndex
```
zsoltk:
```suggestion
internal fun Float.toSegmentProgress(segmentIndex: Int) = (this - segmentIndex).coerceIn(0f, 1f)
```

zsoltk: However, with this solution values `> 1f` will be capped at `1f` and cause extra emissions that are conceptually not correct (if it's out of range, it shouldn't trigger anything).

appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/ui/Interpolator.kt
```diff
@@ -33,28 +30,25 @@ interface Interpolator<NavTarget, ModelState>  {
...
     fun mapKeyframes(
-        keyframes: Keyframes<ModelState>
+        keyframes: Keyframes<ModelState>,
+        segmentIndex: Int
     ): List<FrameModel<NavTarget>> =
         mapSegment(
             keyframes.currentSegment,
-            keyframes.segmentProgress
+            keyframes.getSegmentProgress(segmentIndex),
+            keyframes.progress.toSegmentProgress(segmentIndex)
```
zsoltk, line 48: This call with the current impl can still result in receiving > 1f values?

zsoltk (approved):

zsoltk:
Any way to back this with a test?

appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/Keyframes.kt
antonshilov: Yep, this is exactly why I used filter to stop emitting after reaching 1. I'm not confident if distinctUntilChanged will help in this case because of float comparison

zsoltk (approved):

PR #370, opened 2023-03-03
Added mechanism to query retained instance store

LachlanMcKee:
## Description
Added mechanism to query retained instance store

## Check list

### DIFF
--- a/libraries/core/src/main/kotlin/com/bumble/appyx/core/store/RetainedInstanceStoreImpl.kt
+++ b/libraries/core/src/main/kotlin/com/bumble/appyx/core/store/RetainedInstanceStoreImpl.kt
@@ -5,15 +5,29 @@ internal class RetainedInstanceStoreImpl : RetainedInstanceStore {
     private val map: MutableMap<String, MutableMap<String, ValueHolder<*>>> = HashMap()
 
     @Suppress("UNCHECKED_CAST")
-    override fun <T : Any> get(storeId: String, key: String, disposer: (T) -> Unit, factory: () -> T): T =
+    override fun <T : Any> get(
+        storeId: String,
+        key: String,
+        disposer: (T) -> Unit,
+        factory: () -> T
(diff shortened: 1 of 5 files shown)


CherryPerry:
What is use case?

CherryPerry (approved):

libraries/core/src/main/kotlin/com/bumble/appyx/core/store/RetainedInstanceStoreImpl.kt
```diff
@@ -11,6 +11,22 @@ internal class RetainedInstanceStoreImpl : RetainedInstanceStore {
             .getOrPut(key) { ValueHolder(factory(), disposer) }
             .value as T
 
+    override fun isRetainedByStoreId(storeId: String, value: Any): Boolean =
+        map[storeId]
+            ?.let { valueHolderMap ->
```
KovalevAndrey: [nit] could be a reusable function for both implementations
```
private fun MutableMap<String, ValueHolder<*>.hasValue(value: Any) {
    values.any { valueHolder ->
         valueHolder.value === value
    }
}
```

KovalevAndrey (approved):
LGTM, same question though :)

LachlanMcKee:
Thanks! There is definitely a usecase when gradually migrating to the RetainedInstanceStore where you might want to optionally use the `disposables` plugin for a disposable object. 

The benefit of this new function is that when you don't want the retained object to be used within the `disposables` plugin, you could query whether it exists within the RetainedInstanceStore.

libraries/core/src/main/kotlin/com/bumble/appyx/core/store/RetainedInstanceStoreImpl.kt
LachlanMcKee: Good suggestion. I have applied it with a couple of tweaks

PR #377, opened 2023-03-15
SaveableStateHolder remove state for destroyed elements

### LINKED ISSUE #375: SaveableStateRegistry overflow causes TransactionTooLargeException
How to reproduce:

1. Using sample sanbox app, setup https://github.com/guardian/toolargetool tool
2. Enable "Don't keep activities" in dev settings
3. Minimize app, look at logcat. `androidx.lifecycle.BundlableSavedStateRegistry.key` takes `3.6 KB` 
4. In sanbox navigate back and forth to NavModel Examples
5. Minimize app again, now `androidx.lifecycle.BundlableSavedStateRegistry.key` takes much more than `3.6 KB` 

What's wrong:
If you'll take a look into saved bundle, you can see that even disposed Nodes still keeps some of their data in saved state registry
Thus, more we navigates through the app, more garbage collected in `SavedStateRegistry`, and at some point it causes `TransactionTooLargeException` after app pausing

You can find saved `NavKeys` and some data in output bundle using debugger:
`(outputBundle.getBundle("androidx.lifecycle.BundlableSavedStateRegistry.key").get("SaveableStateRegistry:-1") as BaseBundle)`
Than `mMap -> values[0] -> [0]`

My assumption what's wrong (i might be way off):
Appyx registers `SaveableStateProvider` https://github.com/bumble-tech/appyx/blob/1.x/libraries/core/src/main/kotlin/com/bumble/appyx/core/composable/Child.kt#L46 but never unregisters it, neither Compose does

Want to notice that in sample app overhead after each navigation isn't that big, but in my app (may be because of NavTarget data size) each navigation takes much more memory

KovalevAndrey:
## Description

Fixes https://github.com/bumble-tech/appyx/issues/375

Remove saveableStateHolder state for destroyed elements

## Check list

### DIFF
--- a/libraries/core/src/main/kotlin/com/bumble/appyx/core/composable/Children.kt
+++ b/libraries/core/src/main/kotlin/com/bumble/appyx/core/composable/Children.kt
@@ -127,18 +127,17 @@ class ChildrenTransitionScope<T : Any, S>(
     ) {
         val saveableStateHolder = rememberSaveableStateHolder()
 
-        val removedElementsFlow = remember {
-            navModel.removedElementKeys().map { list ->
-                list
-                    .filter { clazz.isInstance(it.navTarget) }
-            }
-        }
-        LaunchedEffect(this@ChildrenTransitionScope) {
-            removedElementsFlow.collect { deletedElements ->
-                deletedElements.forEach { navKey ->
-                    saveableStateHolder.removeState(navKey)
+        LaunchedEffect(navModel) {
+            navModel
+                .removedElementKeys()
+                .map { list ->
+                    list.filter { clazz.isInstance(it.navTarget) }
                 }
-            }
+                .collect { deletedKeys ->
+                    deletedKeys.forEach { navKey ->
+                        saveableStateHolder.removeState(navKey)
+                    }
+                }
         }
 
         val visibleElementsFlow = remember {
--- a/libraries/core/src/main/kotlin/com/bumble/appyx/core/composable/Children.kt
+++ b/libraries/core/src/main/kotlin/com/bumble/appyx/core/composable/Children.kt
@@ -2,6 +2,7 @@ package com.bumble.appyx.core.composable
 
 import androidx.compose.foundation.layout.Box
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.derivedStateOf
 import androidx.compose.runtime.getValue
@@ -15,13 +16,14 @@ import androidx.compose.ui.layout.onSizeChanged
 import androidx.compose.ui.platform.LocalDensity
 import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.IntSize
-import com.bumble.appyx.core.node.ParentNode
 import com.bumble.appyx.core.navigation.NavModel
+import com.bumble.appyx.core.navigation.removedElementKeys
 import com.bumble.appyx.core.navigation.transition.JumpToEndTransitionHandler
 import com.bumble.appyx.core.navigation.transition.TransitionBounds
 import com.bumble.appyx.core.navigation.transition.TransitionDescriptor
 import com.bumble.appyx.core.navigation.transition.TransitionHandler
 import com.bumble.appyx.core.navigation.transition.TransitionParams
+import com.bumble.appyx.core.node.ParentNode
 import kotlinx.coroutines.flow.map
 import kotlin.reflect.KClass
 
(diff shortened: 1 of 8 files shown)


libraries/core/src/main/kotlin/com/bumble/appyx/core/navigation/NavModelExt.kt
```diff
@@ -0,0 +1,24 @@
...
+
+import com.bumble.appyx.core.withPrevious
+import kotlinx.coroutines.flow.map
+
+fun <NavTarget, State> NavModel<NavTarget, State>.removedElementKeys() =
+    this
+        .elements
+        .withPrevious()
+        .map { values ->
+            val previousKeys = values.previous?.map { it.key }
```
CherryPerry: `?.map { }.orEmpty()` instead of null check.

CherryPerry: You can also use `mapTo(HashSet()) {}`. Then you can subtract one set from another to find delta between them (check child node manager).

libraries/core/src/main/kotlin/com/bumble/appyx/core/navigation/NavModelExt.kt
```diff
@@ -0,0 +1,24 @@
...
+            val currentKeys = values.current.map { it.key }
+            if (previousKeys == null) {
+                emptyList()
+            } else {
+                val deletedElements = mutableListOf<NavKey<NavTarget>>()
+                previousKeys.forEach { element ->
+                    if (currentKeys.contains(element).not()) {
```
CherryPerry: Please, do not use `.not()`

libraries/core/src/main/kotlin/com/bumble/appyx/core/composable/Children.kt
```diff
@@ -123,6 +125,22 @@ class ChildrenTransitionScope<T : Any, S>(
...
     ) {
+        val saveableStateHolder = rememberSaveableStateHolder()
+
+        val removedElementsFlow = remember {
+            navModel.removedElementKeys().map { list ->
+                list
+                    .filter { clazz.isInstance(it.navTarget) }
+            }
+        }
+        LaunchedEffect(this@ChildrenTransitionScope) {
```
CherryPerry: Why `ChildrenTransitionScope` here and not `navModel`?
Probably we can move flow creating here because it will be invoked almost only once (`navModel` won't change).

libraries/core/src/main/kotlin/com/bumble/appyx/core/composable/Children.kt
```diff
@@ -123,6 +125,22 @@ class ChildrenTransitionScope<T : Any, S>(
...
+            removedElementsFlow.collect { deletedElements ->
+                deletedElements.forEach { navKey ->
+                    saveableStateHolder.removeState(navKey)
```
CherryPerry: A little sad to see that we have 2 different places to control it (set key in `Child` and remove key in `Children`), but I can't propose anything better.

libraries/core/src/main/kotlin/com/bumble/appyx/core/navigation/NavModelExt.kt
```diff
@@ -0,0 +1,24 @@
+package com.bumble.appyx.core.navigation
+
+import com.bumble.appyx.core.withPrevious
+import kotlinx.coroutines.flow.map
+
+fun <NavTarget, State> NavModel<NavTarget, State>.removedElementKeys() =
```
CherryPerry: Make `internal`.

libraries/core/src/main/kotlin/com/bumble/appyx/core/composable/Children.kt
KovalevAndrey: > A little sad to see that we have 2 different places to control it (set key in `Child` and remove key in `Children`), but I can't propose anything better.

Yes, in order not to introduce breaking change we have to do it in 2 places

libraries/core/src/main/kotlin/com/bumble/appyx/core/composable/Children.kt
KovalevAndrey: moved

libraries/core/src/main/kotlin/com/bumble/appyx/core/navigation/NavModelExt.kt
KovalevAndrey: simplified it but in a different way

CherryPerry (approved):

PR #382, opened 2023-03-23
1.x update androidx lifecycle

### LINKED ISSUE #376: Update Lifefycle to 2.6.0
Update androix lifecycle to the[ latest stable release ](https://developer.android.com/jetpack/androidx/releases/lifecycle#2.6.0)

https://github.com/bumble-tech/appyx/issues/373

KovalevAndrey:
## Description

Fixes https://github.com/bumble-tech/appyx/issues/376

- Androidx lifecycle updated to 2.6.1
- RIBs updated to 0.38.0 to align Androidx lifecycle version
- Kotlin and compose complier version have to be updated as well to match kotlin version 1.8.10 which is used in Androidx lifecycle
- With kotlin update I had to update AGP as well. 

## Check list


libraries/core/build.gradle.kts
```diff
@@ -45,9 +44,10 @@ dependencies {
     implementation(composeBom)
     implementation(libs.androidx.activity.compose)
     implementation(libs.androidx.lifecycle.java8)
-    implementation(libs.androidx.lifecycle.runtime)
     implementation(libs.compose.foundation.layout)
 
+    runtimeOnly(libs.kotlin.coroutines.android)
```
CherryPerry: Can you explain this change?

CherryPerry (approved):

libraries/core/build.gradle.kts
KovalevAndrey: should be api, reverted

PR #388, opened 2023-03-31
Add more tests

vladcipariu91:
## Description

...

## Check list


appyx-interactions/android/src/androidTest/kotlin/com/bumble/appyx/interactions/sample/testdrive/KeyframesTest.kt
```diff
@@ -53,6 +53,22 @@ class KeyframesTest {
         composeTestRule.snapshot("${javaClass.simpleName}_${nameRule.methodName}")
     }
 
+    @Test
+    fun when_in_a_segment_then_possible_to_enqueue_more_segments() {
```
zsoltk: Let's consider if we could describe our expectation for the screenshots in text (in general), to improve maintainability for others. This test and the screenshot makes all sense now, but it took a bit of effort to put it together why is that what I should be seeing.

zsoltk (approved):

appyx-interactions/android/src/androidTest/kotlin/com/bumble/appyx/interactions/sample/testdrive/KeyframesTest.kt
vladcipariu91: do you think this name is slightly better
when_keyframe_is_running_then_possible_to_enqueue_more_operations

PR #390, opened 2023-04-03
Move clipToBounds from MotionController to Children

zsoltk:
Move `clipToBounds` to UI (`Children`, which then passes it through `UiContext`). Simplifies the model not having to know about this flag, and will further simplify creating properties (next PR).


appyx-interactions/android/src/main/kotlin/com/bumble/appyx/interactions/sample/Children.kt
```diff
@@ -56,7 +57,6 @@ fun <InteractionTarget : Any, ModelState : Any> Children(
         modifier = modifier
             .fillMaxSize()
             .composed {
-                val clipToBounds by interactionModel.clipToBounds.collectAsState()
```
KovalevAndrey, line 59: if we don't collect state anymore we don't need `composed` modifier

KovalevAndrey (approved):

PR #391, opened 2023-04-04
Add ability to define node customisation lazily

LachlanMcKee:
## Description
Currently it is not possible to define a node customisation lazily which means that potentially a large amount of classes are instantiated when they are not needed.

This PR aims to add a new API that allows this to be done lazily.

## Check list

### DIFF
--- a/libraries/customisations/src/main/kotlin/com/bumble/appyx/utils/customisations/LazyMutableNodeCustomisationDirectory.kt
+++ b/libraries/customisations/src/main/kotlin/com/bumble/appyx/utils/customisations/LazyMutableNodeCustomisationDirectory.kt
@@ -1,0 +1,11 @@
+package com.bumble.appyx.utils.customisations
+
+import kotlin.reflect.KClass
+
+@Deprecated("This interface will be merged into MutableNodeCustomisationDirectory as part of 1.2")
+interface LazyMutableNodeCustomisationDirectory: MutableNodeCustomisationDirectory {
+
+    fun <T : Any> putSubDirectoryLazy(key: KClass<T>, valueProvider: () -> NodeCustomisationDirectory)
+
+    fun <T : NodeCustomisation> putLazy(key: KClass<T>, valueProvider: () -> T)
+}
--- a/libraries/customisations/src/main/kotlin/com/bumble/appyx/utils/customisations/NodeCustomisationDirectoryImpl.kt
+++ b/libraries/customisations/src/main/kotlin/com/bumble/appyx/utils/customisations/NodeCustomisationDirectoryImpl.kt
@@ -1,20 +1,37 @@
 package com.bumble.appyx.utils.customisations
 
+import com.bumble.appyx.utils.customisations.NodeCustomisationDirectoryImpl.DirectoryValue.LazyDirectoryValue
+import com.bumble.appyx.utils.customisations.NodeCustomisationDirectoryImpl.DirectoryValue.NonLazyDirectoryValue
 import kotlin.reflect.KClass
 
+@Suppress("TooManyFunctions")
 open class NodeCustomisationDirectoryImpl(
     override val parent: NodeCustomisationDirectory? = null
-) : MutableNodeCustomisationDirectory {
+) : LazyMutableNodeCustomisationDirectory {
 
-    private val map: MutableMap<Any, Any> = hashMapOf()
+    private val map: MutableMap<Any, DirectoryValue> = hashMapOf()
 
+    @Deprecated(
+        "Use putLazy to avoid potential performance issues",
+        replaceWith = ReplaceWith("putLazy(key) { value }")
+    )
     override fun <T : NodeCustomisation> put(key: KClass<T>, value: T) {
-        map[key] = value
+        map[key] = NonLazyDirectoryValue(value)
+    }
+
+    override fun <T : NodeCustomisation> putLazy(key: KClass<T>, valueProvider: () -> T) {
+        map[key] = LazyDirectoryValue(lazy(valueProvider))
     }
 
     fun <T : Any> put(vararg values: T) {
         values.forEach {
-            map[it::class] = it
+            map[it::class] = NonLazyDirectoryValue(it)
+        }
+    }
+
+    fun <T : Any> putLazy(vararg values: Lazy<T>) {
+        values.forEach {
+            map[it::class] = LazyDirectoryValue(it)
         }
     }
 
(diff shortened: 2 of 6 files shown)


libraries/customisations/src/main/kotlin/com/bumble/appyx/utils/customisations/LazyMutableNodeCustomisationDirectory.kt
```diff
@@ -0,0 +1,10 @@
+package com.bumble.appyx.utils.customisations
+
+import kotlin.reflect.KClass
+
+interface LazyMutableNodeCustomisationDirectory: MutableNodeCustomisationDirectory {
```
CherryPerry, line 6: Let's add directly to `MutableNodeCustomisationDirectory` instead?
I see your point of keeping API compatibility, but I highly doubt anybody has ever implemented their own implementation of this class.
As an alternative, we can provide a default implementation of these methods that are invoked instantly to `MutableNodeCustomisationDirectory` interface.

libraries/customisations/src/main/kotlin/com/bumble/appyx/utils/customisations/LazyMutableNodeCustomisationDirectory.kt
```diff
@@ -0,0 +1,10 @@
...
+
+    fun <T : Any> putLazySubDirectory(key: KClass<T>, valueProvider: () -> NodeCustomisationDirectory)
+
+    fun <T : NodeCustomisation> putLazy(key: KClass<T>, valueProvider: () -> T)
```
CherryPerry, line 10: Let's rename it to just `put` and deprecate a regular non-lazy `put`?

zsoltk (approved):

libraries/customisations/src/main/kotlin/com/bumble/appyx/utils/customisations/NodeCustomisationDirectoryImpl.kt
```diff
@@ -32,21 +37,42 @@ open class NodeCustomisationDirectoryImpl(
...
+            NodeCustomisationDirectoryImpl(this@NodeCustomisationDirectoryImpl).apply(block)
+        }
     }
+
+    /**
+     * Gets a value from the map, if the value is lazy the lazy value will be initialized.
+     */
+    private fun <T: Any> MutableMap<Any, Any>.lazyGet(key: Any?): T? =
+        get(key).let { customisation ->
+            if (customisation is Lazy<*>) {
```
CherryPerry: I would say it is kinda of dangerous (1 step more dangerous than it is right now).

Maybe
```kotlin
val map : MutableMap<Any, Holder> = hashMapOf()

sealed class Holder {
  class Initialized(value)
  class Lazy(lamda)
}
```

libraries/customisations/src/main/kotlin/com/bumble/appyx/utils/customisations/LazyMutableNodeCustomisationDirectory.kt
LachlanMcKee: Sounds good

libraries/customisations/src/main/kotlin/com/bumble/appyx/utils/customisations/NodeCustomisationDirectoryImpl.kt
LachlanMcKee: Yeah... I could wrap it with a sealed class instead? Thoughts?

I.e.

Sealed class Item {
     Data class Lazy(Val lazyitem: Any)
     Data class NotLazy(Val item: Any)
}

I also thought about having two different maps, but it got a bit ugly

libraries/customisations/src/main/kotlin/com/bumble/appyx/utils/customisations/LazyMutableNodeCustomisationDirectory.kt
LachlanMcKee: Sure thing. I was thinking about you specifically when I avoided breaking the API compatibility haha.

KovalevAndrey: should we release 1.2 in this case?

LachlanMcKee: Hmm. If you're happy to release 1.2 soon, otherwise I will just leave it as is, and add a comment saying to remove it in 1.2?

I would prefer to release this reasonably soon if possible

KovalevAndrey: if we extend `MutableNodeCustomisationDirectory` we'll break source compatibility. Yes, most likely nobody implemented their own `MutableNodeCustomisationDirectory`  but nevertheless it's source incompatible change. Maybe we can keep it as `LazyMutableNodeCustomisationDirectory` with `Deprecated("Will be merged into MutableNodeCustomisationDirectory in 1.2")` ?

LachlanMcKee: yeah, I will add the deprecated annotation

LachlanMcKee: done @KovalevAndrey

libraries/customisations/src/main/kotlin/com/bumble/appyx/utils/customisations/NodeCustomisationDirectoryImpl.kt
LachlanMcKee: done @CherryPerry

libraries/customisations/src/main/kotlin/com/bumble/appyx/utils/customisations/LazyMutableNodeCustomisationDirectory.kt
LachlanMcKee: @CherryPerry Unfortunately I cannot rename this do do overload ambiguity.

KovalevAndrey (approved):

CherryPerry (approved):

PR #417, opened 2023-04-24
Add utils prefix for artifacts id for all submodules in util module

KovalevAndrey:
## Description

Publish all utils submodules with utils prefix

## Check list


zsoltk (approved):

LachlanMcKee (approved):
This won't scale very well, but we can improve it later.

CherryPerry (approved):

PR #421, opened 2023-04-25
2.x fix publishing pluging

KovalevAndrey:
## Description

Fixes gradle task to work correctly with multiptaform.

1. Previously it was creating one publication for kotlin multiplatfrom module instead of one per target
2. Refactored so most of the code is reused across Android, JavaLibrary and KMP publications
3. Created extension so we can explicitly provide artifactId for publications

Local maven publishing result:

Screenshot 2023-04-27 at 09 25 46

## Check list


appyx-components/demos/common/build.gradle.kts
```diff
@@ -6,6 +6,10 @@ plugins {
     id("appyx-publish-multiplatform")
 }
 
+publishingPlugin {
```
CherryPerry: Why publish demos?

plugins/publish-plugin/src/main/kotlin/AndroidAppyxPublishPlugin.kt
```diff
@@ -14,5 +18,16 @@ internal class AndroidAppyxPublishPlugin : ProjectPlugin() {
         }
     }
 
+    override fun PublicationContainer.createPublications(project: Project) {
+        create<MavenPublication>("appyxRelease") {
```
CherryPerry: Let's sync with MPP naming – `androidRelease`

plugins/publish-plugin/src/main/kotlin/JavaAppyxPublishPlugin.kt
```diff
@@ -10,5 +14,16 @@ internal class JavaAppyxPublishPlugin : ProjectPlugin() {
         }
     }
 
+    override fun PublicationContainer.createPublications(project: Project) {
+        create<MavenPublication>("appyxRelease") {
```
CherryPerry: Let's sync with MPP naming – `jvm`

plugins/publish-plugin/src/main/kotlin/MultiplatformAppyxPublishPlugin.kt
```diff
@@ -1,12 +1,37 @@
...
+        val javadocJar by project.tasks.registering(Jar::class) {
+            archiveClassifier.set("javadoc")
+        }
+
+        val artifactId = project.extensions.getByType(ProjectPluginExtension::class.java).artifactId
+
+        mppExtension.targets.all {
+            mavenPublication {
+                if (artifactId.isNotEmpty()) {
+                    this.artifactId = "$artifactId-$targetName"
```
CherryPerry: I hope it works as expected.

CherryPerry (approved):

plugins/publish-plugin/src/main/kotlin/MultiplatformAppyxPublishPlugin.kt
KovalevAndrey: it does - check the artifacts on the image in the description

appyx-components/demos/common/build.gradle.kts
KovalevAndrey: we want to showcase some stuff but we won't commit to supporting it

CherryPerry (approved):

PR #425, opened 2023-05-03
Up navigation interop

CherryPerry:
## Description

Up navigation from interop node is handled by `Activity` instead of RIB parent.
This PR properly propagate the call to RIB parent.

Unhandled case: invoking `navigateUp` from `onBuilt`, but looks like it does not work anyway in RIBs too (have tried once), so decided to just report exception.

## Check list

### DIFF
--- a/libraries/interop-ribs/src/main/kotlin/com/bumble/appyx/interop/ribs/UpHandlingAppyxIntegrationPoint.kt
+++ b/libraries/interop-ribs/src/main/kotlin/com/bumble/appyx/interop/ribs/UpHandlingAppyxIntegrationPoint.kt
@@ -1,0 +1,32 @@
+package com.bumble.appyx.interop.ribs
+
+import com.badoo.ribs.util.RIBs
+import com.bumble.appyx.core.integrationpoint.IntegrationPoint
+import com.bumble.appyx.core.integrationpoint.activitystarter.ActivityStarter
+import com.bumble.appyx.core.integrationpoint.permissionrequester.PermissionRequester
+
+internal class UpHandlingAppyxIntegrationPoint(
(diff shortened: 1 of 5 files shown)


libraries/interop-ribs/src/main/kotlin/com/bumble/appyx/interop/ribs/UpHandlingAppyxIntegrationPoint.kt
```diff
@@ -0,0 +1,32 @@
+package com.bumble.appyx.interop.ribs
+
+import com.badoo.ribs.util.RIBs
+import com.bumble.appyx.core.integrationpoint.IntegrationPoint
+import com.bumble.appyx.core.integrationpoint.activitystarter.ActivityStarter
+import com.bumble.appyx.core.integrationpoint.permissionrequester.PermissionRequester
+
+internal class UpHandlingAppyxIntegrationPoint(
```
zsoltk, line 8: Would it make sense to add `handleUpNavigation` to the abstract `IntegrationPoint` instead?

zsoltk (approved):

libraries/interop-ribs/src/main/kotlin/com/bumble/appyx/interop/ribs/UpHandlingAppyxIntegrationPoint.kt
CherryPerry: It is already implicitly
```kotlin
abstract class IntegrationPoint: UpNavigationHandler

fun interface UpNavigationHandler {
    fun handleUpNavigation()
}
```

Default implementation uses `Activity.finish()`, but here we need to pass invocation to RIB instead.

PR #431, opened 2023-05-15
Multiplatform utils

### LINKED ISSUE #426: Create common module for multiplatform utilities
First step to expanding multiplatform support.

For Parcelable and other multiplatform core utilities, add a module within utils/multiplatform-common that can be a dependency for any multiplatform module

Ninjars:
## Description

Refactored cross platform Parcelable implementation into a multiplatform utility module, setting it up for re-use in other multiplatform modules.


utils/multiplatform-common/src/commonMain/kotlin/com/bumble/appyx/utils/multiplatform_common/Parcelable.kt
```diff
@@ -0,0 +1,11 @@
+package com.bumble.appyx.utils.multiplatform_common
```
zsoltk: I'd avoid underscores in directory structure & package names (the module name is fine as is). Can we go with either two levels `multiplatform/common/` (and `multiplatform.common` respectively), or just a single one called `multiplatform`?

zsoltk (approved):

PR #444, opened 2023-05-24
Make Position implement Parcelable to avoid crash

KovalevAndrey:
## Description

Position should implement parcelable to avoid crash when saving state

## Check list


zsoltk (approved):
Btw we should enforce this compile-time for all other cases too: #310

PR #450, opened 2023-06-05
Improve gesture handling

manuel-martos:
## Description

...

## Check list


appyx-components/demos/android/src/main/kotlin/com/bumble/appyx/components/demos/cards/DatingCards.kt
```diff
@@ -41,57 +35,43 @@ fun DatingCards(modifier: Modifier = Modifier) {
...
-                                )
-                            }
-                        )
-                    }
+        onDragStart = { position -> cards.onStartDrag(position) },
+        onDrag = { change, dragAmount, _ ->
+            change.consume()
+            cards.onDrag(dragAmount, density)
+            true
+        },
```
zsoltk: This could work automatically in `DraggableChildren` since we already pass the interaction model, which implements the drag related methods via interface.

appyx-components/demos/android/src/main/kotlin/com/bumble/appyx/components/demos/cards/DatingCards.kt
```diff
@@ -41,57 +35,43 @@ fun DatingCards(modifier: Modifier = Modifier) {
...
+        onDragEnd = {
+            cards.onDragEnd(
+                completionThreshold = 0.15f,
+                completeGestureSpec = spring(stiffness = Spring.StiffnessLow),
+                revertGestureSpec = spring(stiffness = Spring.StiffnessMedium)
             )
```
zsoltk: Maybe even this can be moved to automatic code under the hood, and then we only need to pass an optional `GestureSettleConfig` object that holds these 3 params.

appyx-components/internal/common/src/commonMain/kotlin/com/bumble/appyx/components/internal/testdrive/ui/TargetUiState.kt
```diff
@@ -2,11 +2,13 @@ package com.bumble.appyx.components.internal.testdrive.ui
...
 import com.bumble.appyx.interactions.core.ui.property.impl.BackgroundColor
 import com.bumble.appyx.interactions.core.ui.property.impl.Position
+import com.bumble.appyx.interactions.core.ui.property.impl.RotationZ
 import com.bumble.appyx.interactions.core.ui.state.MutableUiStateSpecs
 
 @Suppress("unused")
 @MutableUiStateSpecs
 class TargetUiState(
     val position: Position.Target,
+    val rotationZ: RotationZ.Target,
```
zsoltk: If we add the rotation to the default ui, it will probably mess up the screenshot tests. We can make this the new baseline, or we can make it a separate example.

appyx-components/internal/common/src/commonMain/kotlin/com/bumble/appyx/components/internal/testdrive/TestDriveExperiment.kt
```diff
@@ -133,34 +129,35 @@ fun <InteractionTarget : Any> TestDriveUi(
...
+                    change.consume()
+                    testDrive.onDrag(dragAmount, density)
+                    true
+                } else {
+                    change.consume()
+                    testDrive.onDragEnd()
+                    false
+                }
+            },
+            onDragEnd = { testDrive.onDragEnd() },
```
zsoltk: Similarly to `DatingCards.kt`, I guess this could be automatic too?

appyx-components/demos/android/src/main/kotlin/com/bumble/appyx/components/demos/cards/DatingCards.kt
```diff
@@ -41,57 +35,43 @@ fun DatingCards(modifier: Modifier = Modifier) {
...
     InteractionModelSetup(cards)
 
-    Children(
+    DraggableChildren(
         modifier = modifier
             .fillMaxSize()
             .background(appyx_dark)
             .padding(16.dp),
+        screenWidthPx = (LocalConfiguration.current.screenWidthDp * LocalDensity.current.density).roundToInt(),
+        screenHeightPx = (LocalConfiguration.current.screenHeightDp * LocalDensity.current.density).roundToInt(),
```
zsoltk, line 50: I guess this could be moved one level deeper, so that client code doesn't need to specify it.

zsoltk (approved):

appyx-components/demos/android/src/main/kotlin/com/bumble/appyx/components/demos/cards/DatingCards.kt
manuel-martos: Not easy to move it deeper because it's using `LocalConfiguration` that is Android specific.

appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/gesture/GestureValidator.kt
```diff
@@ -0,0 +1,13 @@
+package com.bumble.appyx.interactions.core.gesture
+
+import androidx.compose.runtime.Stable
+import androidx.compose.ui.geometry.Offset
+import androidx.compose.ui.geometry.Rect
+
+typealias GestureValidator = (Offset, Rect) -> Boolean
```
zsoltk: I'd probably prefer an `interface` over a `typealias`, and implementations rather than top level `val`s.

zsoltk (approved):

PR #454, opened 2023-06-07
Add Padding.kt motion property

KovalevAndrey:
## Description

Add padding motion property to animate padding

## Check list

### DIFF
--- a/appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/ui/property/impl/Padding.kt
+++ b/appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/ui/property/impl/Padding.kt
@@ -1,0 +1,111 @@
+package com.bumble.appyx.interactions.core.ui.property.impl
+
+import androidx.compose.animation.core.Animatable
+import androidx.compose.animation.core.AnimationVector4D
+import androidx.compose.animation.core.Easing
+import androidx.compose.animation.core.TwoWayConverter
+import androidx.compose.foundation.layout.padding
+import androidx.compose.runtime.collectAsState
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.composed
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.lerp
+import com.bumble.appyx.interactions.core.ui.context.UiContext
+import com.bumble.appyx.interactions.core.ui.property.Interpolatable
+import com.bumble.appyx.interactions.core.ui.property.MotionProperty
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+
+class Padding(
+    uiContext: UiContext,
+    target: Target,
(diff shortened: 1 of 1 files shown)


appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/ui/property/impl/Padding.kt
```diff
@@ -0,0 +1,105 @@
...
+    easing = target.easing,
+    visibilityThreshold = visibilityThreshold,
+    displacement = displacement
+), Interpolatable<Padding.Target> {
+
+    class Target(
+        val top: Dp = 0.dp,
+        val bottom: Dp = 0.dp,
+        val start: Dp = 0.dp,
+        val end: Dp = 0.dp,
```
manuel-martos, line 37: This change clashes with auto-generation code. Could you please rollback to PaddingValues?

manuel-martos: Also, it would be nice to enrich PaddingValues with more constructors and operators:
- Single param constructor to setup all values equally.
- Two params constructor to setup horizontal and vertical padding.
- Arithmetic operators: plus, minus, times (optional).

KovalevAndrey: funny that @zsoltk asked to do exactly the opposite, @zsoltk wdyt?

zsoltk: I find it redundant to wrap them + redundant in the naming. The whole class of a `Padding.Target` is already doing exactly that, expressing a padding value (esp. if we can extract the easing from these classes later). 

@manuel-martos is there any way to make it work with the code generation, or is it too much of a hassle?

If we definitely need to keep it, I'd name it just `Value`, nested as a sibling to `Target`, both in `Padding`, to remove the redundancy from naming. So you would have `Padding`, `Padding.Target`, `Padding.Value`.

manuel-martos: The actual issue lies in how the generated code implements the animateTo method. It's something similar to the following Kotlin code:

```kotlin
override suspend fun animateTo(
  scope: CoroutineScope,
  target: TargetUiState,
  springSpec: SpringSpec<Float>,
): Unit {
  listOf(
    scope.async {
      padding.animateTo(
        target.padding.value, // <-- Error: unresolved reference
        spring(springSpec.dampingRatio, springSpec.stiffness),
      )
    },
  ).awaitAll()
}
```

In the KSP processor, we can write a special case for Padding and provide all the resolvable references (`top, bottom, left, right`), but we will still have an issue here because the `animateTo` method from `MotionProperty<T, V>` only accepts a single parameter from the `Target`. So, I believe we still need to keep the redundancy you mentioned.

PR #455, opened 2023-06-08
Gesture improvements

zsoltk:
## Description

- Pass the model state to gesture factory
- Adds a much easier API for drag->progress calculation along an axis
- `TestDriveExperiment` uses the new apis, and now allows for diagonal gestures too
- Removes necessity to always define gesture operations as `KEYFRAME`, it's now added automatically
- Some fixes for possible bugs

### DIFF
--- a/appyx-components/spotlight/common/src/commonMain/kotlin/com/bumble/appyx/components/spotlight/ui/slider/SpotlightSlider.kt
+++ b/appyx-components/spotlight/common/src/commonMain/kotlin/com/bumble/appyx/components/spotlight/ui/slider/SpotlightSlider.kt
@@ -12,7 +12,6 @@ import com.bumble.appyx.components.spotlight.SpotlightModel.State.ElementState.D
 import com.bumble.appyx.components.spotlight.SpotlightModel.State.ElementState.STANDARD
 import com.bumble.appyx.components.spotlight.operation.Next
 import com.bumble.appyx.components.spotlight.operation.Previous
-import com.bumble.appyx.interactions.core.model.transition.Operation.Mode.KEYFRAME
 import com.bumble.appyx.interactions.core.ui.context.TransitionBounds
 import com.bumble.appyx.interactions.core.ui.context.UiContext
 import com.bumble.appyx.interactions.core.ui.gesture.Drag
@@ -104,12 +103,12 @@ class SpotlightSlider<InteractionTarget : Any>(
             Orientation.Horizontal -> {
                 when (dragHorizontalDirection(delta)) {
                     Drag.HorizontalDirection.LEFT -> Gesture(
-                        operation = if (reverseOrientation) Previous(KEYFRAME) else Next(KEYFRAME),
+                        operation = if (reverseOrientation) Previous() else Next(),
                         axis = Offset(-width, 0f)
                     )
 
                     else -> Gesture(
-                        operation = if (reverseOrientation) Next(KEYFRAME) else Previous(KEYFRAME),
+                        operation = if (reverseOrientation) Next() else Previous(),
                         axis = Offset(width, 0f)
                     )
                 }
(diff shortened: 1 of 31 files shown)


appyx-components/spotlight/common/src/commonMain/kotlin/com/bumble/appyx/components/spotlight/ui/slider/SpotlightSlider.kt
```diff
@@ -93,42 +92,39 @@ class SpotlightSlider<InteractionTarget : Any>(
...
                     else ->
                         Gesture(
-                            operation = if (reverseOrientation) Next(KEYFRAME) else Previous(KEYFRAME),
-                            dragToProgress = { offset -> (offset.y / height) },
-                            partial = { offset, partial -> offset.copy(y = partial * height) }
+                            operation = if (reverseOrientation) Next() else Previous(),
+                            axis = Offset(height, 0f)
                         )
                 }
             }
```
manuel-martos: Be careful here, `axis` should update `y` component, not `x` one. Otherwise `SpotlightStack3D` will be broken.

appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/model/progress/DragProgressController.kt
```diff
@@ -35,10 +36,13 @@ class DragProgressController<InteractionTarget : Any, State>(
...
-        gestureFactory().createGesture(dragAmount, density)
         if (_gestureFactory == null) {
             _gestureFactory = { dragAmount ->
-                gestureFactory().createGesture(dragAmount, density)
+                gestureFactory().createGesture(
+                    state = model.output.value.currentTargetState,
+                    delta = dragAmount,
+                    density = density
+                )
             }
```
manuel-martos, line 46: We can tackle in the context of another ticket, but I find it overkill having to instantiate `GestureFactory` + `Gesture` in this way. For instance, pretend that we have spotlight with `activeIndex = 0`. If we swipe to the right meaning that we should move to previous element, as that element doesn't exist, we cancel gesture immediately and `_gestureFactory` becomes `null`. So, as long as the user is swiping to the right, we keep instantiating gesture factories that last for really short time. No UI change, but  is generating a huge memory pressure.

appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/model/progress/DragProgressController.kt
```diff
@@ -51,14 +55,17 @@ class DragProgressController<InteractionTarget : Any, State>(
     private fun consumeDrag(dragAmount: Offset) {
         val currentState = model.output.value
         require(dragAmount.isValid()) { "dragAmount is NaN" }
-        require(dragAmount.getDistance() > 0f) { "dragAmount distance is 0" }
+        if (dragAmount.getDistanceSquared() == 0f) {
+            return
+        }
```
manuel-martos, line 60: 💯

appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/ui/math/Vector.kt
```diff
@@ -0,0 +1,51 @@
...
+            when {
+                x < 0 -> 180 + deg
+                y < 0 -> 360 + deg
+                else -> deg
+            }
+        }
+    }
+
+    return (deg + 90) % 360
+}
```
manuel-martos: If precision isn't an issue here, I would suggest to return `Float` instead of `Double`.

appyx-components/spotlight/common/src/commonMain/kotlin/com/bumble/appyx/components/spotlight/ui/slider/SpotlightSlider.kt
zsoltk: Good catch!

PR #461, opened 2023-06-12
Puzzle 15

vladcipariu91:
## Description

...

## Check list

### DIFF
--- a/appyx-components/demos/common/src/commonMain/kotlin/com/bumble/appyx/components/demos/puzzle15/Puzzle15Model.kt
+++ b/appyx-components/demos/common/src/commonMain/kotlin/com/bumble/appyx/components/demos/puzzle15/Puzzle15Model.kt
@@ -1,0 +1,58 @@
+package com.bumble.appyx.components.demos.puzzle15
+
+import com.bumble.appyx.components.demos.puzzle15.Puzzle15Model.Tile
+import com.bumble.appyx.interactions.Parcelable
+import com.bumble.appyx.interactions.Parcelize
+import com.bumble.appyx.interactions.RawValue
+import com.bumble.appyx.interactions.core.Element
+import com.bumble.appyx.interactions.core.asElement
+import com.bumble.appyx.interactions.core.model.transition.BaseTransitionModel
+import com.bumble.appyx.interactions.core.state.SavedStateMap
+
+class Puzzle15Model(
+    savedStateMap: SavedStateMap?
+) : BaseTransitionModel<Tile, Puzzle15Model.State>(
+    savedStateMap = savedStateMap
+) {
+    data class Tile(val value: String)
+
+    @Parcelize
+    data class State(
+        val items: @RawValue List<Element<Tile>>,
+        val emptyTileIndex: Int
+    ) : Parcelable
+
+    private fun createItems(): List<Tile> {
+        val items = mutableListOf<Tile>()
+        repeat(16) {
+            if (it == 0) {
+                items.add(Tile(""))
+            } else {
+                items.add(Tile(it.toString()))
+            }
+        }
+
+        return items.shuffled()
+    }
+
+    override val initialState: State
+        get() {
+            val items = createItems()
+            val emptyTile = items.find { it.value.isEmpty() }
+            val index = items.indexOf(emptyTile)
+            return State(
+                items = items.map { it.asElement() },
+                emptyTileIndex = index
+            )
+        }
+
+    override fun State.destroyedElements(): Set<Element<Tile>> = emptySet()
+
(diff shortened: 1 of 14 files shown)


appyx-components/demos/common/src/commonMain/kotlin/com/bumble/appyx/components/demos/puzzle15/ui/Puzzle15MotionController.kt
```diff
@@ -0,0 +1,82 @@
...
+) {
+
+    override fun Puzzle15Model.State.toUiTargets(): List<MatchedTargetUiState<Tile, TargetUiState>> =
+        items.mapIndexed { index, tileElements ->
+            MatchedTargetUiState(
+                element = tileElements, // todo: if i create items here this misbehaves because the ids will be different
+                targetUiState = TargetUiState(
+                    position = Position.Target(
+                        value = DpOffset(
+                            x = (index % 4 * 60).dp,
```
zsoltk: I wonder if instead baking in `60`, we could take the total width/height from transition bounds, and just divide them by 4. This would make the entire layout flexible.

appyx-components/demos/common/src/commonMain/kotlin/com/bumble/appyx/components/demos/puzzle15/operation/Swap.kt
```diff
@@ -0,0 +1,73 @@
...
+
+    private fun offsetForDirection(direction: Direction) =
+        when (direction) {
+            Direction.Left -> -1
+            Direction.Up -> -4
+            Direction.Right -> 1
+            Direction.Down -> 4
+        }
+
+    sealed class Direction : Parcelable {
```
zsoltk: Can this be just an `enum class`?

appyx-components/demos/common/src/commonMain/kotlin/com/bumble/appyx/components/demos/puzzle15/Puzzle15Model.kt
```diff
@@ -0,0 +1,58 @@
...
+import com.bumble.appyx.interactions.core.asElement
+import com.bumble.appyx.interactions.core.model.transition.BaseTransitionModel
+import com.bumble.appyx.interactions.core.state.SavedStateMap
+
+class Puzzle15Model(
+    savedStateMap: SavedStateMap?
+) : BaseTransitionModel<Tile, Puzzle15Model.State>(
+    savedStateMap = savedStateMap
+) {
+    data class Tile(val value: String)
```
zsoltk: I'd rather not store it as a `String`. Displaying text in the composition is only one visualisation, it could as well be a slice of a photo. Having it as an `enum` instead could benefit from exhaustive when branching in client code to choose whatever visualisation needed.

settings.gradle.kts
```diff
@@ -60,5 +61,6 @@ project(":appyx-components:backstack:common").name = "backstack"
 project(":appyx-components:spotlight:common").name = "spotlight"
 project(":appyx-components:demos:common").name = "demos"
 project(":appyx-interactions:common").name = "appyx-interactions"
+project(":appyx-components:demos:web").name = "puzzle-15-web"
```
zsoltk: This component isn't necessarily representative of all of the appyx-components demos. Let's move it to its own module inside demos.

vladcipariu91: i renamed the module to `demos-web` instead, it seems too much to have a dedicated module for this game. This now sits at the same level with Promoter and Cards (see appyx-components:demos:common -> commonMain)

demos/appyx-interactions/android/src/main/kotlin/com/bumble/appyx/interactions/MainActivity.kt
```diff
@@ -66,7 +66,7 @@ class MainActivity : ComponentActivity() {
                             Button({ content = 7 }) { Text("7") }
                         }
                         when (content) {
-                            1 -> DatingCards()
+                            1 -> Puzzle15()
```
zsoltk: We still need the `DatingCards`, can you add this as a new item to the bottom of the list?

settings.gradle.kts
zsoltk: It makes it sound like this example is representative of all the possible web demos, that's why I'd rather have it in its own module, and then depend on a bunch of them for a compiled web demo.

demos/appyx-interactions/android/src/main/kotlin/com/bumble/appyx/interactions/MainActivity.kt
vladcipariu91: oops

Issue #464, opened 2023-06-15
MutableUiState visibility is incorrectly resolved in some cases


KovalevAndrey:
If `MutableUiState` has `Scale` or `Rotation` motion properties that modify the bounds of UiElement then visibility will be incorrectly resolved as it doesn't take these properties into account.

New approach should take into account all `MotionProperties` which can modify element bounds, transform bounds based of these values and calculate if the element is visible on the screen.

Example:

The initial bounds are `0, 0` and `x, y`.  Element has `scale = x2`, `rotationY = 90`, `offsetX = -30`, the transformed bounds should be  `-30, 0` and `2y - 30, 2x`.

Final bounds should be matched to transition bounds to calculate if the element is visible or not

PR #474, opened 2023-06-16
Rework visibility approach

KovalevAndrey:
## Description

Fixes https://github.com/bumble-tech/appyx/issues/464

**Old approach:**
Every `MotionProperty` is exposing `isVisibleFlow`, and `BaseMutableUiState` visibility was a combined value from all these flows. The problem with that approach is that each `MotionProperty` doesn't have enough information to calculate its visibility correctly. For instance, `Position` can resolve its visibility correctly but if `Scale` or `Rotation` is present they will not be considered.

**New approach**
In the new approach, I duplicate the modifier which will be used for transitions, and add `onPlaced` modifier which will help to understand where the element bounds relative to the parent (if clipToBounds == true) or root (if clipToBounds == false). The only factor that is not considered is alpha. For that reason I'm allowing `MotionProperties` to override `isVisibleFlow` if they influence visibility and its not related to position.
 

## Check list

### DIFF
--- a/appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/ui/output/ElementUiModel.kt
+++ b/appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/ui/output/ElementUiModel.kt
@@ -9,7 +9,14 @@ import kotlinx.coroutines.flow.StateFlow
 data class ElementUiModel<InteractionTarget>(
     val element: Element<InteractionTarget>,
     val visibleState: StateFlow<Boolean>,
-    val animationContainer: @Composable () -> Unit,
+    /**
+     * The purpose of this container is to be present in composition as long as corresponding
+     * Element exists. With its help we calculate visibility of the element and animate MotionProperties
+     * even if the element is invisible. This help to break circular dependency when element is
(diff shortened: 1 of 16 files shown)


appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/ui/output/ElementUiModel.kt
```diff
@@ -9,7 +9,14 @@ import kotlinx.coroutines.flow.StateFlow
...
     val visibleState: StateFlow<Boolean>,
-    val animationContainer: @Composable () -> Unit,
+    /**
+     * The purpose of this container is to be present in composition as long as corresponding
+     * Element exists. With its help we calculate visibility of the element and animate MotionProperties
+     * even if the element is invisible. This help to break circular dependency when element is
+     * not present in the composition and therefore we can't update MotionProperties, and because
+     * MotionProperties are not updated element can never enter composition even when it should.
+     */
+    val persistedContainer: @Composable () -> Unit,
```
zsoltk:
```suggestion
    val persistentContainer: @Composable () -> Unit,
```

zsoltk (approved):

PR #476, opened 2023-06-19
Avoid consumes all input

### LINKED ISSUE #466: DraggableChildren consumes all input
DraggableChildren composable consumes all pointer events even if it can't process them (e.g. the conditions don't apply). This means no other composables overlapping the same area can take any inputs (a button won't take clicks).

manuel-martos:
## Description

Rework a little bit current approach applying this changes:
- Move `pointerInput` back to elements instead of parent.
- Track to different bounding boxes: transformed + local
- Track offset between bounding boxes centre.
- Wrap `pointerInput` with `offset(center)` and `offset(-center)` to make sure gesture inputs are locally translated to element coordinate system.

## Check list


appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/model/progress/Draggable.kt
```diff
@@ -12,5 +12,5 @@ interface Draggable : HasDefaultAnimationSpec<Float> {
 
     fun onDrag(dragAmount: Offset, density: Density)
 
-    fun onDragEnd()
+    fun onDragEnd(onSettled: (() -> Unit)? = null)
```
zsoltk: The naming couples the concept of dragging to settling, which isn't necessarily true. If we implement free drag, it will not need any settling on the model for example.

zsoltk (approved):

Issue #494, opened 2023-06-28
Allow flag to control whether a gesture can be continuous


zsoltk:
**Situation right now:**

- If a drag gesture can cover more than 1 length of an operation, the remainder is automatically interpreted as another model operation

**Cases where this is desired:**

- Spotlight
- Test drive
- In general: models where a continuous dragging beyond one operation is what we want

**Cases where this is not desired:**

- Back stack pop
- Puzzle 15 move an element

In these cases the gesture should stop at the first operation's 100% completion boundary.

**Idea:**

Allow a flag inside `Gesture` to control this behaviour, and then the gesture factory (client code) can decide.

zsoltk:
Fixed in #599

PR #503, opened 2023-06-30
Settle after gesture release

### LINKED ISSUE #495: Remainder of gesture operation doesn't settle
**Scenario:**

- Drag longer than 100% in any model (e.g. puzzle 15 element drag)

**ER:**

- Even without #494, if the drag gets interpreted as more than one operation, it should settle after the gesture

**AR:**

- The state is left hanging with some amount of drag, it doesn't settle
- As a consequence, no other operations can be applied until manually settled (which might make it feel like it's unresponsive)

manuel-martos:
## Description

Jetpack Compose does not send the `Release` motion event type to the `Element` that is being dragged, which is why the `onDragEnd` event does not get triggered, and no settling occurs. As a workaround, the parent element needs to be made aware of when the `Release` event is triggered and allow the underlying motion controller to properly end the current drag and settle gesture.

## Check list

### DIFF
--- a/appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/DraggableChildren.kt
+++ b/appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/DraggableChildren.kt
@@ -18,6 +18,7 @@ import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clipToBounds
 import androidx.compose.ui.geometry.Offset
 import androidx.compose.ui.geometry.Rect
+import androidx.compose.ui.input.pointer.PointerEventType
 import androidx.compose.ui.input.pointer.pointerInput
 import androidx.compose.ui.layout.boundsInParent
 import androidx.compose.ui.layout.onPlaced
@@ -29,6 +30,7 @@ import com.bumble.appyx.interactions.core.gesture.GestureValidator
 import com.bumble.appyx.interactions.core.gesture.GestureValidator.Companion.defaultValidator
 import com.bumble.appyx.interactions.core.gesture.detectDragGesturesOrCancellation
 import com.bumble.appyx.interactions.core.model.BaseAppyxComponent
+import com.bumble.appyx.interactions.core.modifiers.onPointerEvent
 import com.bumble.appyx.interactions.core.ui.context.TransitionBounds
 import com.bumble.appyx.interactions.core.ui.context.UiContext
 import com.bumble.appyx.interactions.core.ui.output.ElementUiModel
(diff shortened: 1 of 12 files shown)


appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/DraggableChildren.kt
```diff
@@ -82,6 +84,11 @@ fun <InteractionTarget : Any, ModelState : Any> DraggableAppyxComponent(
                     clipToBounds = clipToBounds
                 )
             }
+            .onPointerEvent {
+                if (it.type == PointerEventType.Release && appyxComponent.isDragging()) {
+                    appyxComponent.onDragEnd()
```
KovalevAndrey: maybe we can always report to component with `onRelease()` and check internally if it's being dragged or not. This way we don't need to expose the state

appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/DraggableChildren.kt
```diff
@@ -82,6 +84,11 @@ fun <InteractionTarget : Any, ModelState : Any> DraggableAppyxComponent(
                     clipToBounds = clipToBounds
                 )
             }
+            .onPointerEvent {
```
KovalevAndrey, line 87: do we need this for puzzle15?

appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/model/progress/DragProgressController.kt
```diff
@@ -52,6 +52,8 @@ class DragProgressController<InteractionTarget : Any, State>(
         _gestureFactory = null
     }
 
+    override fun isDragging(): Boolean = _gestureFactory != null
```
KovalevAndrey: let's explicitly set a dedicated variable `isDragging`

KovalevAndrey (approved):

appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/DraggableChildren.kt
manuel-martos: For puzzle15, we can apply proposal from [this issue](https://github.com/bumble-tech/appyx/issues/494) and avoid carrying out gesture to next segment.

KovalevAndrey: yeah but in my mind they are not mutually exclusive

manuel-martos: They aren't, but let's do baby steps as gesture handling is core feature for this project.

PR #522, opened 2023-07-07
Add redirects for 1.x-related links

zsoltk:


.github/workflows/build.yml
```diff
@@ -148,4 +148,5 @@ jobs:
           python-version: '3.x'
       - run: pip install mkdocs-material
       - run: pip install mkdocs-macros-plugin
+      - run: pip install mkdocs-redirects
```
manuel-martos, line 151: We need to add this same line in `.github/workflows/documentation.yml` file as that's the Github Action that will deploy docs after merging.

manuel-martos (approved):

Issue #429, opened 2023-05-12
Implement platform-specific multiplatform navigation interfaces


Ninjars:
In js and desktop modules implement the interfaces needed for multiplatform to function.

Primarily this looks to be PlatformBackHandler, but there may be other convenience or functionally required classes to add too, such as the AndroidNodeHost convenience class which made it easier to create NodeHosts.

Create demos targeting desktop and web platforms that use navigation module in a basic way as a proof of functionality.

PR #531, opened 2023-07-14
429 implement platform specific multiplatform navigation interfaces

Ninjars:
## Description
Addresses issue https://github.com/bumble-tech/appyx/issues/429

- Implemented PlatformBackHandler for desktop and web platforms
- Implemented NodeHost convenience wrappers to simplify creating the Appyx entry point for desktop and web platforms
- Created working versions of navigation demos for both desktop and web, as proof of concept

Note: DatingCardsNode demo node is unsupported as there's not currently a good multiplatform image loading solution. Coil is looking to support multiplatform image loading in 3.0 release which is in active development, so rather than hack around this issue now it makes sense to wait and revisit when support is available. Alternatively we could look at other image loading libraries in the interim.

## Check list


zsoltk:
> DatingCardsNode demo node is unsupported as there's not currently a good multiplatform image loading solution

We could do it without images, the same as in https://bumble-tech.github.io/appyx/components/experimental/

appyx-navigation/common/src/desktopMain/kotlin/com/bumble/appyx/navigation/platform/OnBackPressedCallback.kt
```diff
@@ -0,0 +1,58 @@
...
+
+import java.util.concurrent.CopyOnWriteArrayList
+
+interface Cancellable {
+    /**
+     * Cancel the subscription. This call should be idempotent, making it safe to
+     * call multiple times.
+     */
+    fun cancel()
+}
```
zsoltk, line 11: Minor thing: we could extract this

appyx-navigation/common/src/desktopMain/kotlin/com/bumble/appyx/navigation/platform/OnBackPressedCallback.kt
```diff
@@ -0,0 +1,58 @@
...
+        for (cancellable in cancellables) {
+            cancellable.cancel()
+        }
+    }
+
+    /**
+     * Callback for handling the [OnBackPressedDispatcher.onBackPressed] event.
+     */
+    abstract fun handleOnBackPressed()
+    fun addCancellable(cancellable: Cancellable) {
```
zsoltk:
```suggestion
    abstract fun handleOnBackPressed()
    
    fun addCancellable(cancellable: Cancellable) {
```

demos/appyx-navigation/desktop/src/desktopMain/kotlin/com/bumble/appyx/navigation/Main.kt
```diff
@@ -0,0 +1,85 @@
...
+private fun onKeyEvent(
+    keyEvent: KeyEvent,
+    events: Channel<Events>,
+    coroutineScope: CoroutineScope = CoroutineScope(SupervisorJob() + Dispatchers.Main),
+): Boolean =
+    when {
+        keyEvent.type == KeyEventType.KeyDown && keyEvent.key == Key.Backspace -> {
+            coroutineScope.launch { events.send(Events.OnBackPressed) }
+            true
+        }
```
zsoltk, line 82: We could consider escape instead of backspace maybe? Should it be configurable by client code?

Ninjars: yeah, this is client code configuring it. Just need to send the event.

Everything in "demos" module I'm considering as client code, so the Main.kt for desktop and web can be seen as examples of a "hello world"-esq entrypoint into an Appyx navigation app that displays the multiplatform "ContainerNode" component.

Most of the platform-specific configuration is being done in the DesktopNodeHost and WebNodeHost, which are in the relevant platform navigation library modules for ease of use as I *think* that they are sufficiently generic to be commonly usable, and are also easy enough to copy-paste and modify if not.

appyx-navigation/common/src/commonMain/kotlin/com/bumble/appyx/navigation/composable/Children.kt
```diff
@@ -28,12 +28,13 @@ import gestureModifier
 import kotlin.math.roundToInt
 
 @Composable
-inline fun <reified InteractionTarget : Any, ModelState : Any> ParentNode<InteractionTarget>.Children(
+fun <InteractionTarget : Any, ModelState : Any> Children(
+    parent: ParentNode<InteractionTarget>,
```
KovalevAndrey: for consistency, let's keep it an extension function on ParenNode?

settings.gradle.kts
```diff
@@ -75,5 +79,6 @@ project(":appyx-components:experimental:promoter:common").name = "promoter"
 project(":appyx-components:experimental:puzzle15:common").name = "puzzle15"
 project(":appyx-components:experimental:puzzle15:web").name = "puzzle15-web"
 project(":appyx-components:internal:test-drive:common").name = "test-drive"
+project(":demos:appyx-navigation:web").name = "navigation-web"
```
KovalevAndrey, line 83: We must not touch platform-specific artifacts. Will be generated automatically. We just need to rename common modules so the namings do not clash
```suggestion
project(":demos:appyx-navigation:common").name = "appyx-navigation"
```

KovalevAndrey (changes requested):

appyx-navigation/common/src/commonMain/kotlin/com/bumble/appyx/navigation/composable/Children.kt
Ninjars: it was the extension on ParentNode that broke on JS - looks like there's a limited amount of scope nesting that the JS compilation system can take. That was a real headache to isolate! It causes a runtime crash without any stacktrace, so had to go literally line by line to infer what was going wrong.

There might be a way to do this in a way compatible with the platform but without there being docs or discussion about this it's hard to see how at this point. Though it is sad that it makes the api a little less neat.

settings.gradle.kts
Ninjars: note that this is the demos common module, not the appyx-navigation library - so probably needs to by ```project(":appyx-navigation:common").name = "appyx-navigation"``` for the navigation module.

For the demo module I was looking at what was done for the puzzle15 component, which has both common and web modules defined in there. Not sure what is required...

KovalevAndrey: Since we're not publishing demos, I don't think we need to rename the module

appyx-navigation/common/src/commonMain/kotlin/com/bumble/appyx/navigation/composable/Children.kt
```diff
@@ -28,12 +28,20 @@ import gestureModifier
...
 
 @Composable
-inline fun <reified InteractionTarget : Any, ModelState : Any> ParentNode<InteractionTarget>.Children(
+fun <InteractionTarget : Any, ModelState : Any> ParentNode<InteractionTarget>.Children(
     interactionModel: BaseInteractionModel<InteractionTarget, ModelState>,
     modifier: Modifier = Modifier,
     clipToBounds: Boolean = false,
-    noinline block: @Composable ChildrenTransitionScope<InteractionTarget, ModelState>.() -> Unit = {
-        children { child, elementUiModel ->
+    block: @Composable (ChildrenTransitionScope<InteractionTarget, ModelState>.() -> Unit)? = null
```
KovalevAndrey, line 35: I like it less than default lambda but if we can't make it work I'm okay with it

KovalevAndrey (approved):

appyx-navigation/common/src/commonMain/kotlin/com/bumble/appyx/navigation/composable/Children.kt
Ninjars: we can always revisit later and see if there's a work-around that JS is happy with.

Issue #465, opened 2023-06-15
Position to take alignment parameter


zsoltk:
Right now we can only pass an `Offset` to `Position.Target`.

The idea is to have a second parameter in the form of a standard compose 2D alignment:

```
TopStart, TopCenter, TopEnd, CenterStart, Center, CenterEnd, BottomStart, BottomCenter, BottomEnd
```

We already know the bounds, so we should be able to calculate an `Offset` relative to them (bear in mind, it's not only the container bounds we need to take, but also the element's own dimensions). E.g.:

- `Position.Target(Offset(100.dp, 100.dp))`  (equivalent of current behaviour, default argument is `Alignment.TopStart`)
- `Position.Target(Offset(100.dp, 100.dp), Alignment.TopStart)`  (equivalent of current behaviour)
- `Position.Target(Offset(-100.dp, -100.dp), Alignment.BottomEnd)`
- etc.

A good application of this would be to update both current `TestDrive` `MotionControllers` so that `B`, `C`, `D` states are (0, 0) relative to their corresponding corners. This should allow TestDriveExperiment to work flexibly with different screen / embed sizes.

Another good application can be in `BackStack` / `Spotlight`, not having to use `width` manually.

zsoltk:
Fixed.

PR #539, opened 2023-07-22
Position alignment

zsoltk:
## Description

Allows `Position` to take an `Alignment` that can be animated: #465 

- `Offset` and `Alignment` can be used independently or together too – the latter allows to define positions that are relative to an anchor, e.g. `Position.Target(Alignment.Center, DpOffset(50.dp, 50.dp))`
- `Alignment` can be any `BiasAlignment`, fractional numbers work too (e.g. `0.5f` = `50%` between `Center` and `End`)
- All `TestDrive`-based examples now animate `Alignment` instead of `DpOffset` = they're no longer dependent on the available screen space. 

This PR should also unlock some better shared element transition capabilities (not in the layout sense, but on the level of visual effects experienced), we should build a small demo for it.

I expect screenshot tests to break as e.g. `CenterStart/CenterEnd` will not be exactly where the previous ad-hoc offsets were set.

## Check list


appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/sample/Children.kt
```diff
@@ -57,38 +57,44 @@ fun <InteractionTarget : Any, ModelState : Any> Children(
...
+                    uiContext = UiContext(
+                        coroutineScope = coroutineScope,
+                        transitionBounds = TransitionBounds(
+                            density = density,
+                            widthPx = it.size.width,
+                            heightPx = it.size.height,
+                            screenWidthPx = screenWidthPx,
+                            screenHeightPx = screenHeightPx
+                        ),
+                        boxScope = this@Box,
```
KovalevAndrey: BoxScope is available at the composition phase while onPlaced is called in the layout phase. We save BoxScope and use it in UiContext to avoid double Box

appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/ui/property/impl/Position.kt
```diff
@@ -1,60 +1,171 @@
...
+                alignment = TopStart,
+                offset = offset
+            )
+        )
+
+        constructor(
+            alignment: Alignment
+        ) : this(
+            value = Value(
+                // Compose hides BiasAlignment type for the most commonly used Alignments.
```
KovalevAndrey: maybe we can create our own type to support stronger typization and avoid this?

appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/DraggableChildren.kt
```diff
@@ -63,12 +67,13 @@ fun <InteractionTarget : Any, ModelState : Any> DraggableAppyxComponent(
...
     val gestureExtraTouchAreaPx = with(density) { gestureExtraTouchArea.toPx() }
     var uiContext by remember { mutableStateOf<UiContext?>(null) }
+    var boxScope: BoxScope? = null
 
     LaunchedEffect(uiContext) {
         uiContext?.let { appyxComponent.updateContext(it) }
     }
     Box(
-        modifier = modifier
+        modifier = Modifier
```
manuel-martos: Input param `modifier` isn't used anymore in this context. Is this intended?

manuel-martos (approved):

appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/DraggableChildren.kt
zsoltk: Good catch!

PR #542, opened 2023-07-25
Desktop widget representation

### LINKED ISSUE #529: Restore desktop sample looks
As a side-effect of #505 now the desktop sample is also having a centered look instead of the original slanted one. I suggest it shouldn't depend on the stock `SpotlightStack3D` class but have its own, independent representation. 

cc @manuel-martos

manuel-martos:
## Description

Create specific representation for Desktop Widget sample.

Fixes #529

## Check list


demos/appyx-interactions/desktop/src/desktopMain/kotlin/com/bumble/appyx/interactions/widgets/ui/WidgetsStack3D.kt
```diff
@@ -0,0 +1,84 @@
...
+        zIndex = ZIndex.Target(0f),
+    )
+
+    private val destroyed: TargetUiState = TargetUiState(
+        rotationX = RotationX.Target(0f),
+        position = Position.Target(DpOffset(0.dp, -height)),
+        scale = Scale.Target(0.25f),
+        alpha = Alpha.Target(0f),
+        zIndex = ZIndex.Target(0f),
+    )
```
zsoltk: I suppose these don't affect the rendering since the widget doesn't trigger Spotlight's element state change operations, only scroll change, and then most of the UI is functional in `TargetUiState::toMutableUiState`? In that case we could just remove this as a red herring.

manuel-martos: You're right. I'll just leave a `default` one inside `private companion object` as we still need it for `toUiTargets` method.

zsoltk (approved):

Issue #546, opened 2023-07-25
Build web samples in parallel


zsoltk:
It's noticeable in `gradle build` that all web samples are built sequentially rather than in parallel (and they take rather long with packaging js). We could potentially improve our build times here.

PR #550, opened 2023-07-29
[Build & CI] Avoid building mkdoc modules twice within PR workflow

LachlanMcKee:
## Description
Improve build & ci by:
- avoiding building the mkdoc modules twice

## Check list


build.gradle.kts
```diff
@@ -73,9 +73,27 @@ allprojects {
     }
 }
 
+val buildTask = tasks.register("buildNonMkdocs")
+
 subprojects {
     plugins.apply("release-dependencies-diff-create")
 
+    // Allows avoiding building these modules as part of CI as they are also built for mkdocs.
+    if (!path.startsWith(":demos:mkdocs:")) {
```
LachlanMcKee, line 83: I believe this covers the 4 types of modules that we have. if you can think of any others, please let me know.

.github/workflows/build.yml
```diff
@@ -143,6 +143,16 @@ jobs:
...
+      - uses: actions/setup-java@v3
+        with:
+          distribution: 'zulu'
+          java-version: '11'
+      - uses: gradle/wrapper-validation-action@v1
+      - uses: gradle/gradle-build-action@v2
+        with:
+          cache-read-only: ${{ env.MAIN_BRANCH != 'true' }}
+      - name: Generate distributions
+        run: ./gradlew jsBrowserDistributionMkdocs --continue
```
LachlanMcKee, line 155: I worry that previously the mkdocs were running serially. By calling this Gradle command before calling mkdocs, the modules can hypothetically be run in parallel.

.github/workflows/build.yml
```diff
@@ -143,6 +143,16 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+      - uses: actions/setup-java@v3
+        with:
+          distribution: 'zulu'
+          java-version: '11'
+      - uses: gradle/wrapper-validation-action@v1
+      - uses: gradle/gradle-build-action@v2
```
LachlanMcKee, line 151: It seems that we were not previously benefitting from any caching.

.github/workflows/build.yml
zsoltk: Related issue, hopefully we can close it with this: #546

LachlanMcKee: hopefully. though I've seen some error messages in gradle saying that the task is not correctly configured, so there are a lot of cache misses.

hopefully in the future Jetbrains will fix the task so that the gradle cache is more effective.

manuel-martos (approved):

PR #556, opened 2023-08-05
Ensure detekt is enabled for all modules

LachlanMcKee:
## Description
As part of updating AGP to 8.1 and also adding 'gradle convention plugins' I've noticed that Detekt is not actually enabled for any multiplatform module, and in some cases lint/detekt is not enabled for Android modules.

## Check list


settings.gradle.kts
```diff
@@ -34,7 +35,6 @@ include(
     ":appyx-components:internal:test-drive:common",
     ":appyx-components:experimental:cards:android",
     ":appyx-components:experimental:cards:common",
-    ":appyx-components:experimental:modal:android",
```
LachlanMcKee, line 37: This does not exist

plugins/convention/src/main/kotlin/com/bumble/appyx/multiplatform/MultiplatformConventionPlugin.kt
```diff
@@ -0,0 +1,29 @@
...
+
+class MultiplatformConventionPlugin : Plugin<Project> {
+    override fun apply(project: Project) {
+        project.extensions.create("appyx", MultiplatformConventionExtension::class.java)
+
+        project.plugins.apply("kotlin-multiplatform")
+        project.plugins.apply("org.jetbrains.compose")
+        project.plugins.apply("appyx-detekt")
+
+        project.extensions.configure<DetektExtension> {
```
LachlanMcKee, line 15: If we create an iOS example in the future this list would need to be updated. annoyingly detekt doesn't look at all directories

demos/appyx-navigation/detekt.yml
```diff
@@ -1,4 +0,0 @@
-style:
```
LachlanMcKee: Now done in the root config

appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/ui/math/Vector.kt
```diff
@@ -37,6 +38,7 @@ fun angleDegrees(vector: Offset): Float  {
     return (deg.toFloat() + 90) % 360
 }
 
+@Suppress("MagicNumber")
```
zsoltk, line 41: Maybe it's better to add it on the entire file or to not turn it on at all? I don't find too much value in adding a restriction we need to keep suppressing everywhere.

plugins/convention/src/main/kotlin/com/bumble/appyx/multiplatform/MultiplatformConventionPlugin.kt
zsoltk: I wonder what would be the best place to record this memo, here it will just get lost.

zsoltk (approved):

appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/ui/math/Vector.kt
LachlanMcKee: I think we should only add it to the top of files we don't plan to fix.

I think some of these should be revisited, and constants created. Thoughts?

plugins/convention/src/main/kotlin/com/bumble/appyx/multiplatform/MultiplatformConventionPlugin.kt
LachlanMcKee: Good question. I'm unsure. Do we have a milestone or something for any planned iOS tickets? It could be added there?

appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/ui/math/Vector.kt
zsoltk: I'm not convinced it adds value in most cases. It also won't play well with `MotionController` instances where all numbers are ui properties. 

I'd rather remove the rule (at least in the scope of interactions/components, we can consider keeping it in places we don't expect ui properties to pop up) as it just adds friction 99% of the time, and an unjustifiable magic number is one of the easier things to catch on a code review anyway.

PR #561, opened 2023-08-15
Add and update multiplatform-related documentation

zsoltk:
## Description

- Added a standalone page for Multiplatform info
  - Lifecycle, Parcelize, etc.
  - Node hosts for different platforms
- Updated quick start guide

## Check list

### DIFF
--- a/documentation/navigation/multiplatform.md
+++ b/documentation/navigation/multiplatform.md
@@ -1,0 +1,211 @@
+---
+title: Appyx Navigation – Multiplatform
+---
+
+# Multiplatform
+
+## Supported platforms
+
+![badge-android](https://img.shields.io/badge/platform-android-brightgreen)
+![badge-jvm](https://img.shields.io/badge/platform-jvm-orange)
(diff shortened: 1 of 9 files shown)


documentation/navigation/multiplatform.md
```diff
@@ -0,0 +1,211 @@
...
+    }
+}
+
+```
+
+## Back handling
+
+On Android back events are handled automatically.
+
+In the above [desktop](#desktop) and [web](#web) examples there are a references to an `onKeyEvent` method.
```
Ninjars, line 190: grammar mistake "there are a references" -> "there is a"

documentation/navigation/multiplatform.md
```diff
@@ -0,0 +1,211 @@
...
+
+You can implement as such, and configure the keybinding which `Key` should trigger a back event:
```
Ninjars, line 192: rephrase
"You can configure any `KeyEvent` to trigger a back event via the events Channel. In this example the OnBackPressed event is launched when the backspace key is pressed down."

PR #576, opened 2023-08-31
Removed unused dependency diff plugin

LachlanMcKee:
## Description
Removed the dependency diff plugin I added a while ago. I never finished it, and unfortunately it's now a blocker for AGP 8.0, so it makes sense to remove it.

Perhaps we can find a 3rd party solution in the future.

## Check list


KovalevAndrey (approved):

LachlanMcKee:
I am confident that this UI test timeout has nothing to do with this PR

Issue #534, opened 2023-07-17
Unify composables


zsoltk:
We have:

- appyx-navigation/AppyxComponent
- appyx-interactions/Children
- appyx-interactions/DraggableAppyxComponent

Ideally we should have one.

FYI @KovalevAndrey @manuel-martos

zsoltk:
Done in #467

PR #585, opened 2023-09-08
Fix drag vs align

zsoltk:
## Description

Related to #539 

**Issue that arose in that PR's context**

- `boundsInParent()` gives an offset relative to 0,0 in parent – this offset is what we use to setup the gesture detection area
- `.align()`, if present in the `elementUiModel` modifiers will also affect the gesture detection area's alignment though, and we can't turn it off, because:
- `.align()` can only be set once in compose (or more precisely, the first one takes priority, any subsequent invocations are ignored), which means we couldn't do the same trick as with offset (+offset, gesture, -offset) to temporarily disable alignment
- but without turning it off temporarily the effect is that `boundsInParent()` offset is always relative to 0,0 whereas it should be relative to the alignment point, resulting in a wrong target area being calculated

**Approach that was implemented instead:**

- A `Box` was extracted from `Child` to apply the gesture detection on – this way it can remain unaffected by alignment change

**The problem with the approach**

- If `Child` has any gesture-related code (e.g. click), pointer input handling can't properly reach the `Box` by bubbling up, since it's not a parent to `Child` but a sibling in the hierarchy
- Dragging the `AppyxComponent` would break in this case and no longer do anything

**New solution**

- Revert back to using a single composable
- Detecting if `PositionInside`/`PositionOutside` `MotionProperty` are present
- Fetch their the alignment-related offset (relative to 0,0) and account for it in the offset calculation for the gesture

**Result**

- Click + Drag works properly
- Drag detection area is calculated properly

...

## Check list

### DIFF
--- a/appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/DraggableChildren.kt
+++ b/appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/DraggableChildren.kt
@@ -197,5 +197,5 @@ fun elementOffset(
         it.alignment.align(elementSize, containerSize, layoutDirection)
     } ?: IntOffset.Zero
 
-    return positionInsideOffset - positionOutsideOffset
+    return positionInsideOffset + positionOutsideOffset
 }
--- a/appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/DraggableChildren.kt
+++ b/appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/DraggableChildren.kt
@@ -10,7 +10,6 @@ import androidx.compose.runtime.Composable
 import androidx.compose.runtime.CompositionLocalProvider
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.collectAsState
-import androidx.compose.runtime.derivedStateOf
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.key
 import androidx.compose.runtime.mutableStateOf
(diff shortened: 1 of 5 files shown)


appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/DraggableChildren.kt
```diff
@@ -97,80 +105,97 @@ fun <InteractionTarget : Any, ModelState : Any> DraggableAppyxComponent(
...
+
+
+@Composable
+fun elementOffset(
+    elementSize: IntSize,
+    containerSize: IntSize,
+): IntOffset {
+
+    val positionInside = motionPropertyRenderValue<Value, PositionInside>()
+    val positionOutside = motionPropertyRenderValue<PositionOutside.Value, PositionOutside>()
```
manuel-martos, line 189: What about importing `PositionInside.Value` and `PositionOutside.Value` using aliases to make this code more orthogonal? Something like this:
```kotlin
import com.bumble.appyx.interactions.core.ui.property.impl.position.PositionInside.Value as PositionInsideValue
import com.bumble.appyx.interactions.core.ui.property.impl.position.PositionOutside.Value as PositionOutsideValue
// ...
val positionInside = motionPropertyRenderValue<PositionInsideValue, PositionInside>()
val positionOutside = motionPropertyRenderValue<PositionOutsideValue, PositionOutside>()
```

appyx-navigation/common/src/commonMain/kotlin/com/bumble/appyx/navigation/composable/AppyxComponent.kt
```diff
@@ -132,65 +139,92 @@ class ChildrenTransitionScope<InteractionTarget : Any, NavState : Any>(
...
     }
+
+    @Composable
+    private fun elementOffset(
+        elementSize: IntSize,
+        containerSize: IntSize
+    ): IntOffset {
+
+        val positionInside = motionPropertyRenderValue<Value, PositionInside>()
+        val positionOutside = motionPropertyRenderValue<PositionOutside.Value, PositionOutside>()
```
manuel-martos, line 217: Same could be applied here.

appyx-navigation/common/src/commonMain/kotlin/com/bumble/appyx/navigation/composable/AppyxComponent.kt
```diff
@@ -132,65 +139,92 @@ class ChildrenTransitionScope<InteractionTarget : Any, NavState : Any>(
...
+                                decorator = block
+                            )
+                        }
                     }
                 }
             }
     }
+
+    @Composable
+    private fun elementOffset(
```
manuel-martos, line 211: What about moving this `elementOffset` composable somewhere common but internal?

zsoltk: As part of #534 we can keep only one rather than moving out.

PR #587, opened 2023-09-11
Fix DraggableChildren and rename it to AppyxComponent

### LINKED ISSUE #586: Promoter is broken in interactions
Looks like #585 broke the promoter example in interactions.

- Other samples seem to work
- Promoter works in navigation app, which makes me think it's another issue of `DraggableChildren` vs `AppyxComponent`.

Would also help: #534 #545

KovalevAndrey:
## Description

Fixes https://github.com/bumble-tech/appyx/issues/586

After merging https://github.com/bumble-tech/appyx/pull/585 we should not apply width and height to the children composables. 

The size of the composable should be determined by client's code modifier

## Check list

### DIFF
--- a/documentation/interactions/usage.md
+++ b/documentation/interactions/usage.md
@@ -85,26 +85,6 @@ fun SomeComposable() {
 }
 ```
 
-You can also make use of the `AppyxComponent` composable, if you have components that allow gestures to control them:
-
-```kotlin
-@Composable
-fun SomeComposable() {
-    AppyxComponent(
-        modifier = Modifier,
-        screenWidthPx = (LocalConfiguration.current.screenWidthDp * LocalDensity.current.density).roundToInt(),
-        screenHeightPx = (LocalConfiguration.current.screenHeightDp * LocalDensity.current.density).roundToInt(),
-        appyxComponent = yourComponent,
-        gestureValidator = permissiveValidator,
-    ) { elementUiModel ->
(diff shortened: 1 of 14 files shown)


documentation/interactions/usage.md
```diff
@@ -85,12 +85,12 @@ fun SomeComposable() {
...
 ```
 
-You can also make use of the `DraggableAppyxComponent` composable, if you have components that allow gestures to control them:
+You can also make use of the `AppyxComponent` composable, if you have components that allow gestures to control them:
 
 ```kotlin
 @Composable
 fun SomeComposable() {
-    DraggableAppyxComponent(
+    AppyxComponent(
```
zsoltk: Doesn't really make sense in context since the preceding paragraph (not shown in the diff) is already about `AppyxComponent`. I guess let's just remove this entire block?

zsoltk: Also worth noting for the future that this only documents the **appyx-interactions** usage of it (which is correct, since page in docs is about that) but we're not documenting too much about it in the **appyx-navigation** docs either (just basic usage in https://bumble-tech.github.io/appyx/navigation/quick-start/).

zsoltk (approved):

documentation/interactions/usage.md
KovalevAndrey: done

PR #589, opened 2023-09-12
Fix visibility resolution when item is less then a parent

KovalevAndrey:
## Description

Previously when resolving an item's visibility the assumption was made that the item always matches the parent's bounds. This leads to incorrectly resolving the visibility of elements that do not match the parent's bounds. 

This bug could be observed in the example of the TestDrive. If we push the element off the screen it will still be visible, and therefore present in the composition.

This PR collects the last known size of the element and applies this size to the invisible dummy composable which resolves visibility for that element.

## Check list

### DIFF
--- a/appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/ui/state/BaseMutableUiState.kt
+++ b/appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/ui/state/BaseMutableUiState.kt
@@ -45,24 +45,23 @@ abstract class BaseMutableUiState<TargetUiState>(
         )
     )
 
-    private val sizeChangedModifier: Modifier = Modifier
-        .onSizeChanged { size ->
-            _size.update { size }
-        }
-
-    protected abstract val combinedMotionPropertyModifier: Modifier
-
     val modifier
         get() = combinedMotionPropertyModifier
             .then(sizeChangedModifier)
+
+    protected abstract val combinedMotionPropertyModifier: Modifier
+
+    private val sizeChangedModifier: Modifier = Modifier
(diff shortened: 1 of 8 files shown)


appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/ui/state/BaseMutableUiState.kt
```diff
@@ -39,12 +45,23 @@ abstract class BaseMutableUiState<TargetUiState>(
...
+    val modifier
+        get() = combinedMotionPropertyModifier
+            .then(sizeChangedModifier)
 
     private val _isBoundsVisible = MutableStateFlow(false)
     private val visibilitySources: Iterable<StateFlow<Boolean>> =
         motionProperties.mapNotNull { it.isVisibleFlow } + _isBoundsVisible
 
+    protected val _size = MutableStateFlow(IntSize.Zero)
+
```
zsoltk: Ordering them like this reads somewhat better, in the order of being referenced:

```suggestion
    val modifier
        get() = combinedMotionPropertyModifier
            .then(sizeChangedModifier)

    protected abstract val combinedMotionPropertyModifier: Modifier
                
    private val sizeChangedModifier: Modifier = Modifier
        .onSizeChanged { size ->
            _size.update { size }
        }
        
    protected val _size = MutableStateFlow(IntSize.Zero)
    private val _isBoundsVisible = MutableStateFlow(false)
    private val visibilitySources: Iterable<StateFlow<Boolean>> =
        motionProperties.mapNotNull { it.isVisibleFlow } + _isBoundsVisible
```

zsoltk (approved):

PR #599, opened 2023-09-18
Issue #494: Add flag to control gesture continuity

vladcipariu91:
[Issue](https://github.com/bumble-tech/appyx/issues/494)

## Description

Added `isContinuous` Boolean flag to `Gesture`. This is set to `true` by default. However in cases where a new gesture being created with the remaining offset isn't desirable this can be set to false.

## Check list


appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/ui/gesture/Gesture.kt
```diff
@@ -6,16 +6,24 @@ import com.bumble.appyx.interactions.core.ui.math.proportionOf
 
 open class Gesture<InteractionTarget, ModelState> internal constructor(
     val operation: Operation<ModelState>,
+    val isContinuous: Boolean,
```
zsoltk: What do you think if we moved it to `GestureFactory`? That way it wouldn't need to be specified per every gesture. The cases we have so far are either all of them are continuous or all of them are not.

vladcipariu91: the current option gives more control, however your suggestion is simpler to use, I would choose the existing one, but it's your call

zsoltk: Can be a combination:
- `GestureFactory` provides the default
- `Gesture` falls back to factory default but allows overriding if you want

zsoltk: I'm assuming based on the cases so far that it's either/or for the whole of it, so if it's only one choice I'd go with factory-level definition for its ease.

vladcipariu91: i'll update to only use on the factory

vladcipariu91: @zsoltk done!

zsoltk (approved):

Issue #604, opened 2023-09-27
Adding a PermanentNavModel with a NavTarget can result in duplicate NavModels


LachlanMcKee:
When passing in a `PermanentNavModel` to a Node similar to:

```
 : ParentNode<StatefulNode1.NavTarget>(
    buildContext = buildContext,
    navModel = PermanentNavModel(
        navTargets = setOf(NavTarget.Child(BASE_COUNTER)), // this is the important issue
        savedStateMap = buildContext.savedStateMap
    )
)
```

It is possible to have multiple PermanentNavModel within `ParentNode.navModel`. This is because `ParentNode` already contains a `permanentNavModel` itself.

When a user calls `PermanentChild` they end up adding this to the `ParentNode.permanentNavModel` rather than the instance they passed in via the constructor.

This results in two Nodes potentially being created, one that is visible and one that is never visible (as it is not rendered)

LachlanMcKee:
This also means that at the moment the benefit of adding a `PermanentNavModel` with an existing `navTargets` cannot 'prewarm' the state, meaning that there is a flicker when the Node is added to the graph due to the `LaunchedEffect` when adding the unique key.

PR #605, opened 2023-09-29
Reuse PermanentNavModel if provided in the list of navModels

KovalevAndrey:
## Description

Fixes: https://github.com/bumble-tech/appyx/issues/604

Problem: If `PermanentNavModel` was provided to the constructor of the ParentNode it will be ignored as ParenNode internally uses its own instance of `PermanentNavModel`. This PR ensures that we reuse the same instance of `PermanentNavModel` if it's provided 

## Check list

### DIFF
--- a/libraries/core/src/main/kotlin/com/bumble/appyx/core/navigation/model/combined/CombinedNavModel.kt
+++ b/libraries/core/src/main/kotlin/com/bumble/appyx/core/navigation/model/combined/CombinedNavModel.kt
@@ -6,6 +6,7 @@ import com.bumble.appyx.core.navigation.NavElements
 import com.bumble.appyx.core.navigation.NavKey
 import com.bumble.appyx.core.navigation.NavModel
 import com.bumble.appyx.core.navigation.NavModelAdapter
+import com.bumble.appyx.core.navigation.model.permanent.PermanentNavModel
 import com.bumble.appyx.core.plugin.Destroyable
 import com.bumble.appyx.core.state.MutableSavedStateMap
 import kotlinx.coroutines.CoroutineScope
@@ -17,6 +18,13 @@ import kotlin.coroutines.EmptyCoroutineContext
 class CombinedNavModel<NavTarget>(
     val navModels: List<NavModel<NavTarget, *>>,
 ) : NavModel<NavTarget, Any?>, Destroyable {
+
+    init {
+        val permanentNavModelCount = navModels.filterIsInstance<PermanentNavModel<*>>().count()
+        check(permanentNavModelCount <= MAX_PERMANENT_NAV_MODEL_COUNT) {
+            "Do not provide more than one PermanentNavModel"
+        }
+    }
 
     constructor(vararg navModels: NavModel<NavTarget, *>) : this(navModels.toList())
 
@@ -59,6 +67,10 @@ class CombinedNavModel<NavTarget>(
     override fun destroy() {
         scope.cancel()
         navModels.filterIsInstance<Destroyable>().forEach { it.destroy() }
+    }
+
+    companion object {
+        private const val MAX_PERMANENT_NAV_MODEL_COUNT = 1
     }
 
 }
--- a/libraries/core/src/main/kotlin/com/bumble/appyx/core/navigation/model/combined/CombinedNavModel.kt
+++ b/libraries/core/src/main/kotlin/com/bumble/appyx/core/navigation/model/combined/CombinedNavModel.kt
@@ -20,9 +20,14 @@ class CombinedNavModel<NavTarget>(
 ) : NavModel<NavTarget, Any?>, Destroyable {
 
     init {
-        val permanentNavModelCount = navModels.filterIsInstance<PermanentNavModel<*>>().count()
+        val permanentNavModelCount = navModels.count { it is PermanentNavModel<*> }
         check(permanentNavModelCount <= MAX_PERMANENT_NAV_MODEL_COUNT) {
-            "Do not provide more than one PermanentNavModel"
+            "CombinedNavModel does not support more than one PermanentNavModel"
+        }
+
+        val hasNoNestedCombinedNavModel = navModels.count { it is CombinedNavModel<*> } == 0
+        check(hasNoNestedCombinedNavModel) {
+            "CombinedNavModel does not support nested CombinedNavModel"
         }
     }
 
--- a/libraries/core/src/main/kotlin/com/bumble/appyx/core/node/ParentNode.kt
+++ b/libraries/core/src/main/kotlin/com/bumble/appyx/core/node/ParentNode.kt
@@ -64,7 +64,7 @@ abstract class ParentNode<NavTarget : Any>(
 ), Resolver<NavTarget> {
 
     private lateinit var permanentNavModel: PermanentNavModel<NavTarget>
-    private var isPermanentNavModelProvided: Boolean? = null
+    private var isPermanentNavModelCreated: Boolean = false
     val navModel: NavModel<NavTarget, *> = resolveNavModels(navModel, buildContext)
 
     /**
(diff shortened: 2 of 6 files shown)


libraries/core/src/main/kotlin/com/bumble/appyx/core/node/ParentNode.kt
```diff
@@ -62,11 +63,52 @@ abstract class ParentNode<NavTarget : Any>(
...
+            navModel + permanentNavModel
+        }
+    }
+
+    private fun retrievePermanentNavModel(navModel: NavModel<NavTarget, *>): PermanentNavModel<NavTarget>? =
+        when (navModel) {
+            is CombinedNavModel<NavTarget> -> {
+                navModel
+                    .navModels
+                    .find { it is PermanentNavModel<NavTarget> } as PermanentNavModel<NavTarget>?
```
LachlanMcKee, line 97: It seems like it's possible (though maybe not advised?) that a developer could nest a `CombinedNavModel` within a `CombinedNavModel`.

Perhaps this case should be covered as well? (if so, then the init block for the CombinedNavModel might need to be updated).

If we don't want to allow nesting `CombinedNavModel` within `CombinedNavModel` perhaps the `init` block of `CombinedNavModel` could prevent this as well? (though this hypothetically could be a breaking change.

i.e.

```
CombinedNavModel(
    listOf(
         CombinedNavModel(
             listOf(
                 PermanentNavModel()
             )
         )
    )
)
```

libraries/core/src/main/kotlin/com/bumble/appyx/core/navigation/model/combined/CombinedNavModel.kt
```diff
@@ -18,6 +19,13 @@ class CombinedNavModel<NavTarget>(
     val navModels: List<NavModel<NavTarget, *>>,
 ) : NavModel<NavTarget, Any?>, Destroyable {
 
+    init {
+        val permanentNavModelCount = navModels.filterIsInstance<PermanentNavModel<*>>().count()
```
LachlanMcKee: Slightly more efficient (as it doesn't create, then discard a list):
`val permanentNavModelCount = navModels.count { it is PermanentNavModel<*> }`

libraries/core/src/main/kotlin/com/bumble/appyx/core/navigation/model/combined/CombinedNavModel.kt
```diff
@@ -18,6 +19,13 @@ class CombinedNavModel<NavTarget>(
...
+        check(permanentNavModelCount <= MAX_PERMANENT_NAV_MODEL_COUNT) {
+            "Do not provide more than one PermanentNavModel"
```
LachlanMcKee: maybe reword to: `CombinedNavModel does not support more than one PermanentNavModel`?

libraries/core/src/main/kotlin/com/bumble/appyx/core/node/ParentNode.kt
```diff
@@ -62,11 +63,52 @@ abstract class ParentNode<NavTarget : Any>(
     plugins = plugins + navModel + childAware
 ), Resolver<NavTarget> {
 
-    private val permanentNavModel = PermanentNavModel<NavTarget>(
-        savedStateMap = buildContext.savedStateMap,
-        key = KEY_PERMANENT_NAV_MODEL,
-    )
-    val navModel: NavModel<NavTarget, *> = permanentNavModel + navModel
+    private lateinit var permanentNavModel: PermanentNavModel<NavTarget>
+    private var isPermanentNavModelProvided: Boolean? = null
```
LachlanMcKee: Does this need to be nullable? perhaps it could be `lateinit` as well?

libraries/core/src/main/kotlin/com/bumble/appyx/core/node/ParentNode.kt
```diff
@@ -62,11 +63,52 @@ abstract class ParentNode<NavTarget : Any>(
...
+            }
+            is PermanentNavModel<NavTarget> -> navModel
+            else -> null
+        }
+
+    private fun createPermanentNavModel(buildContext: BuildContext): PermanentNavModel<NavTarget> =
+        PermanentNavModel(
+            savedStateMap = buildContext.savedStateMap,
+            key = KEY_PERMANENT_NAV_MODEL,
+        )
```
LachlanMcKee: This seems to be unused. was the intention to call this from `resolveNavModels`?

KovalevAndrey: yeah, detekt already complained

libraries/core/src/main/kotlin/com/bumble/appyx/core/node/ParentNode.kt
KovalevAndrey: primitives can't be used with lateinit

CherryPerry: Let's drop `lateinit` instead?

```kotlin
val navModel: NavModel<NavTarget, *>
private val permanentNavModel: PermanentNavModel<NavTarget>
private val isPermanentNavModelProvided: Boolean

init {
   val (combinedNavModel, permanentNavModel, isProvided) = resolveNavModels(navModel, buildContext)
   this.navModel = combinedNavModel
   this.permanentNavModel = permanentNavModel
   isPermanentNavModelProvided = isProvided
}
```

libraries/core/src/main/kotlin/com/bumble/appyx/core/node/ParentNode.kt
KovalevAndrey: yeah, I think we can prevent nesting of `CombinedNavModel` as it doesn't make much sense and harder to support

CherryPerry (commented):
In general OK, but IMHO we are fixing the wrong thing. The library is `Model-driven navigation`, but `@Composable fun PermanentChild(key)` does not fit really into this category, making "navigation" UI-driven. I think we just should provide more convenient `BagNavModel` (that we already use in tests) where we can freely use `add` and `remove` operations and have a Composable function to render its children one by one (by key as it is right now). We should not have such weird shortcuts for child hosting as we have right now and then write a lot of code around handling them.

libraries/core/src/main/kotlin/com/bumble/appyx/core/node/ParentNode.kt
CherryPerry: Nesting makes sense at the current moment.
First you pass `CombinedNavModel` via `ParentNode(navModel = backStack1 + backStack2)` then this `CombinedNavModel` is merged with internal `PermanentNavModel` into another `CombinedNavModel`. So at the end we have `CombinedNavModel` inside `CombinedNavModel` is a pretty basic case.

LachlanMcKee: @KovalevAndrey it sounds like this might be wise to keep? otherwise you should definitely update the changelog to call this out as a breaking change.

KovalevAndrey: @LachlanMcKee Looking at the code I'm leaning towards removing the `PermanentNavModel` from parentNode at all and use it the same way as any other navModels

KovalevAndrey: > Nesting makes sense at the current moment. First you pass `CombinedNavModel` via `ParentNode(navModel = backStack1 + backStack2)` then this `CombinedNavModel` is merged with internal `PermanentNavModel` into another `CombinedNavModel`. So at the end we have `CombinedNavModel` inside `CombinedNavModel` is a pretty basic case.

If we're using the provided API it doesn't get nested

Screenshot 2023-09-30 at 10 48 19

PR #606, opened 2023-10-02
Do not create permanentNavModel inside ParentNode. Provide it via constructor to ParentNode

KovalevAndrey:
## Description

Fixes: https://github.com/bumble-tech/appyx/issues/604

Using `PermanentChild` composable function, we're delaying adding a child to the `ParentNode` until `PermanentChild`  is invoked. This can lead to many problems as the appropriate `Node` will be created only when composable is invoked. Additionally, there's a lot of [complexity](https://github.com/bumble-tech/appyx/pull/605) associated with making sure that we're using only the same instance of `PermanentNavModel` in the cases when one is provided to the constructor.

This PR removes internally created `PermanentNavModel` and forces users to provide it from outside

## Check list

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Pending changes
 
--
+- [#606](https://github.com/bumble-tech/appyx/pull/606) – **Fixed**: Do not create permanentNavModel inside ParentNode. Provide it via constructor to ParentNode
 
 ---
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Pending changes
 
-- [#606](https://github.com/bumble-tech/appyx/pull/606) – **Fixed**: Do not create permanentNavModel inside ParentNode. Provide it via constructor to ParentNode
+- [#606](https://github.com/bumble-tech/appyx/pull/606) – **Breaking change**: Do not create permanentNavModel inside ParentNode. Provide it via constructor to ParentNode
 
 ---
 
--- a/libraries/core/src/androidTest/kotlin/com/bumble/appyx/core/node/PermanentChildTest.kt
+++ b/libraries/core/src/androidTest/kotlin/com/bumble/appyx/core/node/PermanentChildTest.kt
@@ -39,6 +39,13 @@ class PermanentChildTest {
     }
 
     @Test
+    fun `WHEN_permanent_model_does_not_contain_relevant_nav_key_THEN_permanent_child_is_not_rendered`() {
+        rule.start()
+
+        rule.onNode(hasTestTag(TestParentNode.NavTarget::class.java.name)).assertDoesNotExist()
+    }
+
+    @Test
     fun `WHEN_visibility_switched_THEN_permanent_child_is_reused`() {
         createPermanentNavModelWithNavKey()
         rule.start()
--- a/libraries/core/src/androidTest/kotlin/com/bumble/appyx/core/node/PermanentChildTest.kt
+++ b/libraries/core/src/androidTest/kotlin/com/bumble/appyx/core/node/PermanentChildTest.kt
@@ -11,8 +11,9 @@ import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.test.hasTestTag
 import com.bumble.appyx.core.AppyxTestScenario
 import com.bumble.appyx.core.children.nodeOrNull
+import com.bumble.appyx.core.composable.PermanentChild
 import com.bumble.appyx.core.modality.BuildContext
-import com.bumble.appyx.core.navigation.EmptyNavModel
+import com.bumble.appyx.core.navigation.model.permanent.PermanentNavModel
 import kotlinx.parcelize.Parcelize
 import org.junit.Assert.assertEquals
 import org.junit.Rule
@@ -20,20 +21,26 @@ import org.junit.Test
 
 class PermanentChildTest {
 
+    var nodeFactory: (buildContext: BuildContext) -> TestParentNode = {
+        TestParentNode(buildContext = it)
+    }
+
     @get:Rule
     val rule = AppyxTestScenario { buildContext ->
-        TestParentNode(buildContext)
+        nodeFactory(buildContext)
     }
 
     @Test
-    fun permanent_child_is_rendered() {
+    fun `WHEN_permanent_model_contains_relevant_nav_key_THEN_permanent_child_is_rendered`() {
+        createPermanentNavModelWithNavKey()
         rule.start()
 
         rule.onNode(hasTestTag(TestParentNode.NavTarget::class.java.name)).assertExists()
     }
 
     @Test
-    fun permanent_child_is_reused_when_visibility_switched() {
+    fun `WHEN_visibility_switched_THEN_permanent_child_is_reused`() {
+        createPermanentNavModelWithNavKey()
         rule.start()
         rule.node.renderPermanentChild = false
         val childNodes = rule.node.children.value.values.map { it.nodeOrNull }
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## Pending changes
 
--
+- [#606](https://github.com/bumble-tech/appyx/pull/606) – **Fixed**: Do not create permanentNavModel inside ParentNode. Provide it via constructor to ParentNode
```
LachlanMcKee: Is it worth mentioning that the `PermanentChild` now requires a `permanentNavModel` to be provided as a parameter?

libraries/core/src/main/kotlin/com/bumble/appyx/core/composable/PermanentChild.kt
```diff
@@ -0,0 +1,56 @@
...
+import com.bumble.appyx.core.node.Node
+import com.bumble.appyx.core.node.ParentNode
+
+@Composable
+fun <NavTarget : Any> ParentNode<NavTarget>.PermanentChild(
+    permanentNavModel: PermanentNavModel<NavTarget>,
+    navTarget: NavTarget,
+    decorator: @Composable (child: ChildRenderer) -> Unit
+) {
+    val child = remember(navTarget) {
```
LachlanMcKee: You access `permanentNavModel` in this block. should it be part of the `remember`?

libraries/core/src/main/kotlin/com/bumble/appyx/core/composable/PermanentChild.kt
```diff
@@ -0,0 +1,56 @@
...
+        permanentNavModel
+            .elements
+            .value
+            .find { it.key.navTarget == navTarget }
+            ?.let { childOrCreate(it.key) }
+            ?: throw IllegalStateException(
```
LachlanMcKee: Will this break the code within `StatefulNode1.kt`?

Just so I understand, it's now mandatory to pass in a `NavTarget` to the `PermanentNavModel` constructor?

Perhaps we should update `PermanentNavModel` to throw an `IllegalStateException` if there are no `NavTarget`'s provided? Could we also consider removing the vararg then, as this will definitely lead to issues

samples/app/src/main/kotlin/com/bumble/appyx/app/node/slideshow/slide/app/StatefulNode1.kt
```diff
@@ -25,26 +25,28 @@ import androidx.compose.ui.unit.dp
...
 @ExperimentalUnitApi
 @ExperimentalAnimationApi
 @ExperimentalComposeUiApi
 class StatefulNode1(
-    buildContext: BuildContext
-) : ParentNode<StatefulNode1.NavTarget>(
-    buildContext = buildContext,
-    navModel = PermanentNavModel(
+    buildContext: BuildContext,
+    private val permanentNavModel: PermanentNavModel<NavTarget> = PermanentNavModel(
```
LachlanMcKee, line 44: As mentioned above, is it okay that the NavTargets are not passed in here?

CHANGELOG.md
KovalevAndrey: I'd rather mention it in the documentation. We don't have one for permanent child

libraries/core/src/main/kotlin/com/bumble/appyx/core/composable/PermanentChild.kt
KovalevAndrey: good catch

CHANGELOG.md
CherryPerry: Mark as `**Breaking change**`, not just `**Fixed**`.

libraries/core/src/main/kotlin/com/bumble/appyx/core/composable/PermanentChild.kt
CherryPerry: I think we should just ignore if there is no child of the specific type. Otherwise we break functionality where you add a child later. Or maybe provide some optional renderer function.

> Just so I understand, it's now mandatory to pass in a NavTarget to the PermanentNavModel constructor?

Why it should be so? We still have `add`/`addUnique` operations to add nodes later.

KovalevAndrey: I feel like if you have a logic of adding a child later at some point it shouldn't be a part of `PermanentNavModel`

LachlanMcKee: perhaps leave it as is for now, and consider changing this behaviour in 2.0?

CherryPerry: But is not it the case Lachlan wants? Load data and show (add) permanent child only then with the loaded data?

KovalevAndrey: I can remove this check it so it works with `addUnique`  operation

KovalevAndrey: > But is not it the case Lachlan wants? Load data and show (add) permanent child only then with the loaded data?

I don't think so. If he doesn't add a key to `PermanentNavModel` associated node is not created, and therefore doesn't start loading its content

LachlanMcKee: @CherryPerry my usecase is that I have a PermanentChild that is added immediately (via the constructor) and is only shown once the data is loaded (unfortunately in my usecase the ParentNode is responsible for when to show the PermanentChild based on Output of whether the state is loaded or not)

Without this change, the PermanentChild would not correctly start the business logic when added via the PermanentChild constructor (well actually it can, but it renders the wrong one at the moment)

CherryPerry: Anyway, we are breaking existing use case without providing alternative, when you want to add your permanent child later after some flag is triggered. Now you can via `if (myCheck) PermanentChild(NavKey.AddMe)`.

KovalevAndrey: I reverted it back so we can add keys later

libraries/core/src/androidTest/kotlin/com/bumble/appyx/core/node/PermanentChildTest.kt
```diff
@@ -11,29 +11,43 @@ import androidx.compose.ui.platform.testTag
...
-    fun permanent_child_is_rendered() {
+    fun `WHEN_permanent_model_contains_relevant_nav_key_THEN_permanent_child_is_rendered`() {
+        createPermanentNavModelWithNavKey()
         rule.start()
 
         rule.onNode(hasTestTag(TestParentNode.NavTarget::class.java.name)).assertExists()
     }
 
     @Test
-    fun permanent_child_is_reused_when_visibility_switched() {
```
LachlanMcKee, line 36: Can you re-add this test (without createPermanentNavModelWithNavKey)? 
This would verify that the late creation of the child continues to work

KovalevAndrey: The test is still there, i just renamed it to `WHEN_visibility_switched_THEN_permanent_child_is_reused`

libraries/core/src/main/kotlin/com/bumble/appyx/core/composable/PermanentChild.kt
KovalevAndrey: Or we can observe `permanentNavModel` in composition to render permanent children

```
val isAddMeExists = permanentNavModel
    .elements.
    .mapState(scope, SharingStarted.WhileSubscribed()) { navElements ->
         navElements
             .find { it.key.navTarget == navTarget } ?: false
     }
     .collectAsState()

if (isAddMeExists) { PermanentChild() } 

```

otherwise it's difficult to track issues where you forgot to add a key

CherryPerry (approved):

libraries/core/src/main/kotlin/com/bumble/appyx/core/composable/PermanentChild.kt
KovalevAndrey: Perhaps, can be added to an overloaded PermanentChild while default one fails

LachlanMcKee (approved):

PR #609, opened 2023-10-04
Change agent to mac-os latest

KovalevAndrey:
## Description

Release action can not publish iOS targets since the agent is not a macOS. We need to change agents to support that.

## Check list


.github/workflows/build.yml
```diff
@@ -14,7 +14,7 @@ env:
 jobs:
   build:
     name: Build
-    runs-on: ubuntu-latest
+    runs-on: macOS-latest
```
CherryPerry: There is a limit on mac-os agents per repo, makes sense to check of we can merge with instrumentation tests (both variants). Definitely makes sense to merge with `Publication verification`, both builds all libs, the second one just does simple job on packing and posting to local folder. It will reduce amount of unique caches. But fine to do as a separate PR.

.github/workflows/build.yml
```diff
@@ -162,7 +162,7 @@ jobs:
 
   check-documentation:
     name: Check documentation
-    runs-on: ubuntu-latest
+    runs-on: macOS-latest
```
CherryPerry: Documentation builds only JS, it is fine to stay on ubuntu.

.github/workflows/documentation.yml
```diff
@@ -9,7 +9,7 @@ on:
 jobs:
   deploy:
     name: Deploy
-    runs-on: ubuntu-latest
+    runs-on: macOS-latest
```
CherryPerry: Documentation builds only JS, it is fine to stay on ubuntu.
Also this one does not have `Generate distributions` step from check.

gradle/libs.versions.toml
```diff
@@ -3,22 +3,22 @@
...
-composeCompiler = "1.4.4"
+composeCompiler = "1.5.3"
 coroutines = "1.6.4"
 dependencyAnalysis = "1.13.1"
 detekt = "1.21.0"
 junit5 = "5.8.2"
 jvmTarget = "11"
-kotlin = "1.8.10"
-ksp = "1.8.0-1.0.8"
+kotlin = "1.9.10"
```
CherryPerry: Too early, we might not be able to use this version internally.

.github/workflows/build.yml
LachlanMcKee: I would suggest instead separately into even more granular jobs instead.
i.e. a job that builds:
- Android (ubuntu)
- Web (ubuntu)
- Desktop (ubuntu)
- iOS (mac)

or at the very least split iOS build into its own job instead. This will reduce the amount of agents used overall by parallel PRs as the iOS build job would finish a lot faster.

gradle/libs.versions.toml
KovalevAndrey: I know, was testing a fix for an issue https://youtrack.jetbrains.com/issue/KT-60230/Native-unknown-options-iossimulatorversionmin-sdkversion-with-Xcode-15-beta-3 which was fixed in 1.9.10

CherryPerry: You can set xcode version to the one that is currently supported (idk which one on 1.8, just try)
https://github.com/badoo/Reaktive/blob/master/.github/workflows/build.yml#L27

.github/workflows/build.yml
CherryPerry: > I would suggest instead separately into even more granular jobs instead.

It is such a pain to set up and support it 😭 either specifying manually all platform related tasks, or relying on filtering plugins ([trust me](https://medium.com/bumble-tech/continuous-delivery-for-your-kotlin-multiplatform-library-3ab5ad5cba59), but this particular approach does not work anymore because of Gradle updates).

If you are ready to do it, will support you!

CherryPerry (approved):

PR #564, opened 2023-08-16
Replace offset values with alignment

KovalevAndrey:
## Description

Fixes https://github.com/bumble-tech/appyx/issues/562

- Implement custom InsideAlignment and OutsideAlignment
- Replace raw DpOffset with Alignments 

## Check list

- [x] I have updated `CHANGELOG.md` if required.
- [x] I have updated documentation if required.

PR #617, opened 2023-10-08
Unify inside & outside position alignment

zsoltk:
## Description

**Related issue**

- #564 created a separate class for `PositionOutside`, meaning we ended up with two classes that could affect alignment: One for inside the container, one for offsetting whole screen distances for scrolling.

**Problems**

- `.align()` can only be applied once in a `Modifier` change, only the first one having effect
- Following from this, we could have either outside alignment or inside alignment, but not both
- This blocks implementing a shared element transition from a list (outside alignment) to a detail screen (hero element changes alignment inside the screen)
- Adding outside alignment bias into `Position` isn't possible since it already maxed out 4D for its AnimationVector (2 for alignment bias, 2 for relative offset coordinates), a hard limit imposed by Compose.

**Approach in this PR**

- Extract out the relative offset to a standalone property: `PositionOffset` which only takes a `DpOffset`
- This frees up 2 dimensions in the AnimationVector that we can use for outside alignment instead
- Merge `PositionInside` and `PositionOutside` 2+2 alignment bias values into a single property `PositionAlignment`
- The new property can account for both inside and outside alignment in a single modifier correctly

**Migration steps for client code**

In any client code `TargetUiState` instances:

- Replace usages of `PositionInside` and `PositionOutside` with `PositionAlignment`
- Wherever an extra `offset` was also used next to alignments, also introduce a new field for `PositionOffset` and move the value there

## Check list


appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/ui/property/impl/position/PositionAlignment.kt
```diff
@@ -0,0 +1,182 @@
...
+            ),
+        )
+
+    override val modifier: Modifier
+        get() = Modifier.composed {
+            val value = renderValueFlow.collectAsState()
+            val boxScope = requireNotNull(LocalBoxScope.current)
+            with(boxScope) {
+                this@composed
+                    .align(value.value)
```
KovalevAndrey, line 145:
```suggestion
            val value = renderValueFlow.collectAsState()
            val boxScope by requireNotNull(LocalBoxScope.current)
            with(boxScope) {
                this@composed
                    .align(value)
```

KovalevAndrey (approved):

PR #624, opened 2023-10-22
Make renderValueFlow backed by state and remove some of the composed modifiers

KovalevAndrey:
## Description

Getting rid of unnecessary `composed` modifiers to optimize performance   

List of findings during this migration:

1. It matters where you create a Modifier. If a modifier is created outside of the composition state reads are not registered and will not be even when the modifier enters the composition, therefore it will not react to state changes
2. Given the statement above we should return a lambda which returns a modifier instead.  

## Check list


appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/ui/property/MotionProperty.kt
```diff
@@ -48,16 +51,24 @@ abstract class MotionProperty<T, V : AnimationVector>(
 
     private var lastVelocity = animatable.velocity
 
+    /**
+     * Render-ready value that contains applied displacements on top of the [internalValue].
+     * Backed by compose androidx.compose.runtime.State
+     */
+    var renderValue: T by mutableStateOf(animatable.value)
```
KovalevAndrey, line 58: I had to move it up in the class due to https://github.com/bumble-tech/appyx/actions/runs/6606332442/job/17942423218#step:8:284

The flow was updating `renderValue` before it was initialised

PR #631, opened 2023-11-06
Use artifact id only for Android and Java library publications. 

KovalevAndrey:
## Description

Artifact id is used only for Android and Java library project publications. On a multiplatfom project it has not effect. 
In this PR:
1. Removed artifact id plugin extention for multiplatform projects. 
2. Added a check that Kotlin multiplatform project's name is not common.

## Check list


mapm14 (approved):

plugins/publish-plugin/src/main/kotlin/MultiplatformAppyxPublishPlugin.kt
```diff
@@ -7,20 +6,33 @@ import org.gradle.api.publish.maven.MavenPublication
...
+    override fun apply(project: Project) {
+        super.apply(project)
+        project.gradle.projectsEvaluated {
+            if (project.name == "common") {
+                throw IllegalArgumentException(
+                    "Multiplatform project with path ${project.projectDir} should not have its name as common." +
+                            " This will make artifacts rewrite each other." +
+                            " Please change the name of the project in settings.gradle"
+                )
+            }
```
zsoltk, line 18: Nice! 👍 

Maybe any advice in the exception message what naming pattern to apply (or where to find the pattern)?

zsoltk (approved):

plugins/publish-plugin/src/main/kotlin/MultiplatformAppyxPublishPlugin.kt
KovalevAndrey: Naming is unique per project, probably throwing an error when having common as a name should suffice

PR #638, opened 2023-11-27
Fix visibility issue for bottom element in BackStackParallax

aashay-gaikwad:
## Description

Multiple children are visible at the same time in BackStackParallax as the bottom element is still considered in composition. Reason being the bottom element was not out of bounds and had properties making it invisible.

Fixed the issue by setting the bottom element alpha to 0 when not visible (i.e. top element has fully covered it). 

## Check list

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


mapm14 (approved):
Great work! 👏🏻

PR #644, opened 2023-12-18
Refactor AppyxComponent and application of draggable modifier

KovalevAndrey:
## Description

1. AppyxComponent in `appyx-interaction` now takes lambda for drawing children composable. Transition container is applied inside so we clients don't need to apply it themselves. Lambda has only one parameter `Element` instead of `ElementUiModel`
2. AppyxComponent in `appyx-navigation` takes lamda for decorating children composable UI. The UI itself is taken from Nodes.  Lambda has only one parameter `Element` instead of `ElementUiModel` so we don't expose more than we should 
3. AppyxComponent how takes `isGestureBoundingBoxTransformed` which takes into consideration rotation of the element when applying draggable modifier.  
4. AppyxComponent considers if gestures are provided when applying gesture modifier

## Check list


appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/AppyxComponent.kt
```diff
@@ -194,17 +162,114 @@ fun <InteractionTarget : Any, ModelState : Any> AppyxComponent(
...
+            )
+        }
+        .then(elementUiModel.modifier)
+    ) {
+        child(elementUiModel.element)
+    }
+}
+
+@Composable
+private fun <InteractionTarget : Any, ModelState : Any> ChildGestureSupportTransformBoundingBox(
```
KovalevAndrey: At this point I'm not even sure if we need it

appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/ui/gesture/GestureSettleConfig.kt
```diff
@@ -1,12 +1,15 @@
...
-import com.bumble.appyx.interactions.core.ui.helper.DefaultAnimationSpec
 
 @Immutable
 data class GestureSettleConfig(
     val completionThreshold: Float = 0.5f,
     val completeGestureSpec: AnimationSpec<Float> = DefaultAnimationSpec,
     val revertGestureSpec: AnimationSpec<Float> = DefaultAnimationSpec,
 )
+
+val DefaultAnimationSpec: SpringSpec<Float> = spring()
```
KovalevAndrey: place it in the Default file?

appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/core/AppyxInteractionsContainer.kt
```diff
@@ -50,6 +50,11 @@ import com.bumble.appyx.interactions.core.ui.property.motionPropertyRenderValue
 
 private val defaultExtraTouch = 48f.dp
 
+enum class GesturesRelation {
```
zsoltk: 1. I'd extract this to a standalone file in the gestures package
2. I'd rename to `GesturesReferencePoint`
3. I'd provide some comments on what they do

zsoltk: Also for consistency, probably use singular rather than plural, the same as everywhere else

zsoltk (approved):

PR #649, opened 2024-01-09
Don't use Compose leaked internal AtomicReference

### LINKED ISSUE #645: Use of androidx.compose.runtime.AtomicReference
Hi,

Just a heads-up: in the next release of Compose Multiplatform, `androidx.compose.runtime.AtomicReference`, which you are using [here](https://github.com/bumble-tech/appyx/blob/2.x/appyx-navigation/common/src/jsMain/kotlin/com/bumble/appyx/navigation/platform/OnBackPressedCallback.kt) will become `internal`. It was made public accidentally, and we're fixing it now.

You can implement an expect/actual yourself, or use [atomicfu](https://github.com/Kotlin/kotlinx-atomicfu).

manuel-martos:
## Description

In the upcoming Compose Multiplatform version, the leaked AtomicReference class will no longer be exposed. Consequently, it is advisable to discontinue its usage and seek alternative methods to achieve the desired functionality.

## Check list

### DIFF
--- a/appyx-navigation/common/build.gradle.kts
+++ b/appyx-navigation/common/build.gradle.kts
@@ -54,7 +54,6 @@ kotlin {
 
                 implementation(libs.androidx.activity.compose)
                 implementation(libs.androidx.lifecycle.java8)
-
             }
         }
         val desktopMain by getting {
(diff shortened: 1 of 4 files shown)


manuel-martos:
Fixes #645

PR #650, opened 2024-01-10
Remove dead code

manuel-martos:
## Description

This PR includes the following changes:

1. Removed the `:appyx-components:composables` module as it is neither necessary nor referenced in the `settings.gradle` file.

2. Addressed an issue related to Instrumentation tests on CI. The recent increase in the Gradle max heap size from 2G to 4G significantly accelerated the compilation time for the `build` task by a factor of 2x. However, this improvement had a negative impact on the `Instrumentation tests` due to the increased memory footprint required for launching the emulator. To mitigate this, we have decided to revert the Gradle heap size from 4G to 2G specifically for this task.

3. Implemented minor fixes, including the removal of unused imports, addition of missing package statements, and syntax improvements where necessary.

## Check list


.github/workflows/build.yml
```diff
@@ -116,6 +116,7 @@ jobs:
           -Dorg.gradle.workers.max=1
           -Pandroid.testoptions.manageddevices.emulator.gpu="swiftshader_indirect"
           -Pandroid.experimental.testOptions.managedDevices.emulator.showKernelLogging=true
+          -Dorg.gradle.jvmargs=-Xmx2048m
```
zsoltk, line 119: I don't mind, but could also be a separate PR with a reasoning in the description.

zsoltk (approved):

PR #657, opened 2024-01-16
Rename ParentNode & Node to Node and LeafNode

zsoltk:
## Description

- `ParentNode` is now simply called `Node`
- Previous `Node` class is now called `LeafNode`, but it's only a convenience class, and one can simply use `Node` in its place too. It does not add functionality just makes it easier to instantiate `Node` if you don't need children.

The updated migration guide [shows the diff](https://github.com/zsoltk/appyx/blob/17a7dc6b0f2a7e0d7d04d912bd782f6303402c49/documentation/releases/2.0.0-alpha10.md).

## Check list


appyx-navigation/common/src/commonMain/kotlin/com/bumble/appyx/navigation/node/LeafNode.kt
```diff
@@ -0,0 +1,22 @@
...
+import com.bumble.appyx.interactions.core.plugin.Plugin
+import com.bumble.appyx.interactions.permanent.PermanentAppyxComponent
+import com.bumble.appyx.navigation.modality.NodeContext
+
+open class LeafNode(
+    nodeContext: NodeContext,
+    view: NodeView = EmptyNodeView(),
+    plugins: List<Plugin> = listOf(),
+) : Node<Nothing>(
+    appyxComponent = PermanentAppyxComponent(
```
KovalevAndrey: this means that this Node can have children? Yes, we don't have access to this model but is it conceptually correct?

zsoltk: good catch, replacing with `EmptyAppyxComponent`

KovalevAndrey (approved):

PR #667, opened 2024-01-26
Appyx 2.x - Rename integration point

manuel-martos:
## Checklist

- [ ] I've updated `CHANGELOG.md` if required.
- [ ] I've updated the documentation if required.

### DIFF
--- a/documentation/2.x/migrationguide.md
+++ b/documentation/2.x/migrationguide.md
@@ -80,27 +80,6 @@ Artifacts have a `utils-` prefix:
 ```
 
 
-### MainActivity
-
-```diff
--import com.bumble.appyx.core.integration.NodeHost
--import com.bumble.appyx.core.integrationpoint.NodeActivity
-
-+import com.bumble.appyx.navigation.integration.NodeHost
-+import com.bumble.appyx.navigation.integration.NodeActivity
-
-    class MainActivity : NodeActivity() {
-         super.onCreate(savedInstanceState)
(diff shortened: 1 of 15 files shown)


documentation/2.x/migrationguide.md
```diff
@@ -80,27 +80,6 @@ Artifacts have a `utils-` prefix:
...
-         setContent {
-             HelloAppyxTheme {
--                NodeHost(integrationPoint = appyxIntegrationPoint) {
-+                NodeHost(integrationPoint = appyxV2IntegrationPoint) {
-                     RootNode(it)
-                 }
-             }
-     }
-```
-
```
zsoltk, line 103: Maybe it's better if instead of removing this block we update it. After merging #668, let's make a new release in the scope of 1.x

Then we could update this above block to say something like:

> "If you migrate all of your codebase from **1.x** to **2.x** in a single go, you should skip this step. However, if you intend to keep them in parallel for a gradual migration, you must do the following:
>
> 1. Update your 1.x version to [release containing #668]
> 2. Rename `appyxIntegrationPoint` usage in **1.x** code to `appyxV1IntegrationPoint`:
>
> (include the above code but for 1.x)

zsoltk: Let's do this as a standalone PR for documentation when we have the 1.x version out. Merging this ticket now to close API.

zsoltk (approved):

PR #668, opened 2024-01-26
Appyx 1.x - Rename integration point

manuel-martos:
## Checklist

- [ ] I've updated `CHANGELOG.md` if required.
- [ ] I've updated the documentation if required.


zsoltk:
I wonder if we should merge this directly into 1.x or to some 1.x-interop-2.x?

CC @KovalevAndrey

KovalevAndrey:
> I wonder if we should merge this directly into 1.x or to some 1.x-interop-2.x?
> 
> CC @KovalevAndrey

@manuel-martos I think we should merge it into 1.x, it will be much easier to update

KovalevAndrey (approved):

PR #670, opened 2024-01-26
Fix iOS lifecycle

MikePT28:
## Description

Currently, the app's iOS lifecycle wasn't working as expected, when the app was going to the background the Appyx Nodes were not updating their lifecycle states as expected.

This PR solves the issue by making the iOS app to pass the app Lifecycle to the iOSNodeHost.

## Checklist

### DIFF
--- a/appyx-navigation/common/src/iosMain/kotlin/com/bumble/appyx/navigation/platform/LifecycleListener.kt
+++ b/appyx-navigation/common/src/iosMain/kotlin/com/bumble/appyx/navigation/platform/LifecycleListener.kt
@@ -1,0 +1,70 @@
+package com.bumble.appyx.navigation.platform
+
+import com.bumble.appyx.navigation.lifecycle.Lifecycle
+import platform.Foundation.NSNotificationCenter
+import platform.Foundation.NSOperationQueue
+import platform.UIKit.UIApplicationDidBecomeActiveNotification
+import platform.UIKit.UIApplicationDidEnterBackgroundNotification
+import platform.UIKit.UIApplicationWillResignActiveNotification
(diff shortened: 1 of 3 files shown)


mapm14 (approved):

appyx-navigation/common/src/iosMain/kotlin/com/bumble/appyx/navigation/platform/LifecycleListener.kt
```diff
@@ -0,0 +1,73 @@
...
+    private lateinit var didBecomeActiveNotificationObserver: NSObjectProtocol
+    private lateinit var willTerminateNotificationObserver: NSObjectProtocol
+
+    init {
+        created()
+        startObserving()
+    }
+
+    private fun startObserving() {
+        // GOES TO BACKGROUND
```
mapm14: Please don't use full caps for the comments 😄

mapm14 (approved):

appyx-navigation/common/src/iosMain/kotlin/com/bumble/appyx/navigation/platform/LifecycleListener.kt
MikePT28: Done

himanshumistri:
It there way to use latest branch, I have tried `appyx = "2.0.0-alpha10"` but that did not bring this PR #670 changes into the IosNodeHost.kt file.

himanshumistri:
I tried with appyx ="2.0.1" but in Ios Simulator, I am not getting any lifecycle event call.

Issue #672, opened 2024-01-26
Flickering


zsoltk:
Same code as in #671 

Additional issue: wild flickering when navigating for no apparent reason.

https://github.com/bumble-tech/appyx/assets/238198/0f3c925e-6bf8-444a-a841-901e812925fe

PR #643, opened 2023-12-14
Unify AppyxComponent composables

KovalevAndrey:
## Description

AppyxComponent in appyx-interactions and appyx-navigation are identical in their core and should reused as much code as possible to simplify support

## Check list

- [x] I have updated `CHANGELOG.md` if required.
- [x] I have updated documentation if required.

PR #684, opened 2024-02-15
Fix regressions: Local UI state is not persisted and flickering

### LINKED ISSUE #671: Regression: Local UI state is not persisted
## Report

Based on the discussion: https://kotlinlang.slack.com/archives/C05FPNA6P27/p1706194295165759

## STR

1. See the provided projects below
2. Observe the debug numbers or increase the counter with the button on the Profile screen
3. Change navigation in a non-destructive way (spotlight prev/next, or back stack push), come back to previous screen

**Expected**: numbers should be the same
**Actual:** numbers are different

## Investigation so far

- It is possible to persist the state if it's lifted out from the UI into the `Node`, so the `Node` instance is definitely kept alive. It's the local UI state (defined by client code) that's lost.
- The original code didn't use `rememberSaveable`, but I added it in my fork, and it still doesn't work
- When I rolled back the reproducer to use **2.0.0-alpha09** the issue is resolved
- However, the same mechanism for us works in [this demo](https://github.com/bumble-tech/appyx/blob/f5cc89fbdf9a168a70f8e9f11313e6524914179a/demos/sandbox-appyx-navigation/common/src/commonMain/kotlin/com/bumble/appyx/demos/sandbox/navigation/node/backstack/BackStackNode.kt#L76) even with **2.0.0-alpha10**
- I couldn't find any evident differences other than our sample is multiplatform, while the reproducer is pure Android, but that shouldn't be an issue at all.

Any help with further investigation is appreciated.

## Temporary remedy

It is possible to persist the state if it's lifted out from the UI into the `Node`. See the Slack thread for an example.

## Reproducer

I forked the provided sample project to: https://github.com/zsoltk/AppyxNavigationBug

I pushed two branches to it:

- [test-2.0.0-alpha09](https://github.com/zsoltk/AppyxNavigationBug/tree/test-2.0.0-alpha09) (issue is gone)
- [test-2.0.0-alpha10](https://github.com/zsoltk/AppyxNavigationBug/tree/test-2.0.0-alpha10) (issue is present)

KovalevAndrey:
## Description

Fixes https://github.com/bumble-tech/appyx/issues/671 https://github.com/bumble-tech/appyx/issues/672

[In this change](https://github.com/bumble-tech/appyx/pull/643/files#diff-9a0f08f4db634420e9298649dee9f9900f9b1e53fe4cb04bd6901af57e076ad8L22) introduced a bug where I accidentally started saving the state of the transition Box along with children UI state which was wrong. This PR reverts it back and now I only save the state of children UI and not the wrapping Box with transitions.

I checked out the repo in the issue and the bug seems to have gone:

https://github.com/bumble-tech/appyx/assets/5773436/9b502934-6651-4206-889f-b738880ada8e

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 - [#670](https://github.com/bumble-tech/appyx/pull/670) - Fixes ios lifecycle
 - [#673](https://github.com/bumble-tech/appyx/pull/673) – Fix canHandeBackPress typo
+- [#671](https://github.com/bumble-tech/appyx/issue/671) – Fix ui state saving issue
 
 ### Enhancement
 
--- a/appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/composable/AppyxInteractionsContainer.kt
+++ b/appyx-interactions/common/src/commonMain/kotlin/com/bumble/appyx/interactions/composable/AppyxInteractionsContainer.kt
@@ -16,6 +16,7 @@ import androidx.compose.runtime.key
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.rememberCoroutineScope
+import androidx.compose.runtime.saveable.SaveableStateHolder
 import androidx.compose.runtime.saveable.rememberSaveableStateHolder
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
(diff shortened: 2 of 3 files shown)


mmartosdev (approved):

zsoltk (approved):
Approving fix, but please provide better description of 1. what was wrong and 2. why the changes fix the situation.

zsoltk:
Let's also reference #672 that if I understand correctly this also fixes along with #671 (let's provide an updated video capture where appropriate).

zsoltk (approved):

PR #701, opened 2024-04-26
1.x shared element transition and movable content integration

KovalevAndrey:
## Description

Share element transition integration into Appyx. 

1. Clients need to use `withSharedElementTransition: Boolean = true` on `Children` layout when drawing children.
2. Use `sharedElement` Modifier In Node's UI implementation and provide a key that matches the appropriate element's key in the other Node's UI. 

Added support for movableContent. Now we can "move one part of the composable tree" from one child node to another without losing its state. We can only move content between the Node that is transitioning to invisible state to the Node that is transitioning to visible state because movable content is meant to be composed only once at any time. For instance: `NodeA` is currently shown in a `BackStack` and we push `NodeB` to `BackStack`. In this case we can move content from `NodeA` to `NodeB`. To move the content we need to use `localMovableContentWithTargetVisibility` with the correct key in each child Node.

**Do not merge unit the [crash](https://issuetracker.google.com/issues/333679741) is released in alpha08** 

https://github.com/bumble-tech/appyx/assets/5773436/ce5a3def-68b9-4bfc-ac05-6de14080e7e1

## Checklist


libraries/core/src/main/kotlin/com/bumble/appyx/core/composable/Children.kt
```diff
@@ -50,18 +58,42 @@ inline fun <reified NavTarget : Any, State> ParentNode<NavTarget>.Children(
             )
         }
     }
-    Box(modifier = modifier
-        .onSizeChanged {
-            transitionBounds = it
+    if (withSharedElementTransition) {
```
manuelvicnt, line 64: I assume `withSharedElementTransition` won't change dynamically that often but how the API is set up, it _could_. Should we wrap the repeated code (i.e. `CompositionLocalProvider(...) { block(...) }`) with the `moveableContentOf` API to 1) avoid repeating this code and 2) save state if `withSharedElementTransition` were to change?

libraries/core/src/main/kotlin/com/bumble/appyx/core/composable/Children.kt
```diff
@@ -50,18 +58,42 @@ inline fun <reified NavTarget : Any, State> ParentNode<NavTarget>.Children(
...
+        SharedTransitionLayout(modifier = modifier
+            .onSizeChanged {
+                transitionBounds = it
+            }
+        ) {
+            CompositionLocalProvider(
+                LocalSharedElementScope provides this
```
manuelvicnt: Should we document why we're providing `this` vs `null`? is it required?

libraries/core/src/main/kotlin/com/bumble/appyx/core/navigation/NavModelAdapter.kt
```diff
@@ -9,6 +9,7 @@ interface NavModelAdapter<NavTarget, State> {
     val screenState: StateFlow<ScreenState<NavTarget, out State>>
 
     data class ScreenState<NavTarget, State>(
+        val targetStateVisible: NavElements<NavTarget, out State> = emptyList(),
```
manuelvicnt: Can we add more info about what this `targetStateVisible` represents? It's a bit out of context in the `ScreenState` class. 

`onScreen` and `offScreen` makes sense but not `targetStateVisible`. How's that state supposed to mutate/evolve over time?

libraries/core/src/main/kotlin/com/bumble/appyx/core/composable/Children.kt
```diff
@@ -140,30 +175,32 @@ class ChildrenTransitionScope<T : Any, S>(
...
-                    transitionHandler,
-                    block
-                )
+        children
+            .onScreen
+            .filter { clazz.isInstance(it.key.navTarget) }
+            .forEach { navElement ->
+                key(navElement.key.id) {
+                    CompositionLocalProvider(
+                        LocalNodeTargetVisibility provides
```
manuelvicnt, line 201: Are there any tests covering this? Should we add some of them? Feels like the `ScreenState` and a few other things could benefit from them? Thoughts?

libraries/core/src/main/kotlin/com/bumble/appyx/core/navigation/NavModelAdapter.kt
KovalevAndrey: yeah, it conceptually belongs to the `onScreen` element. We need to know which elements from the `onScreen` list are transitioning to `visibleTargetState` but it will break the API if I make a wrapper class

libraries/core/src/main/kotlin/com/bumble/appyx/core/composable/Children.kt
KovalevAndrey: I can extract repeatable parts to a separate composable but it doesn't make sense to you `moveableContentOf` the layout either supports shared element transition or not. I can't come up with a case when we need dynamically change it. WDYT?

libraries/core/src/main/kotlin/com/bumble/appyx/core/composable/Children.kt
KovalevAndrey: I can document, yes. Null is required because if we don't pass it `LocalSharedElementScope` can still be consumed  on grand children level even if their parent doesn't support SharedElement transition

libraries/core/src/main/kotlin/com/bumble/appyx/core/composable/Children.kt
KovalevAndrey: I can add a UI test for that

libraries/core/src/main/kotlin/com/bumble/appyx/core/navigation/NavModelAdapter.kt
manuelvicnt: Gotcha! Can we be more explicit then? `onScreenVisibleTargetState` perhaps?

libraries/core/src/main/kotlin/com/bumble/appyx/core/composable/Children.kt
manuelvicnt: The use case could be that an app allows the user to enable/disable animations on the fly. It's definitely an edge case and something that wouldn't happen very often. But given this is an open source Architecture library and this affects the public API, I guess people could use it for _anything_. If I change `withSharedElementTransition` on the fly, I wouldn't want to lose state.

libraries/core/src/main/kotlin/com/bumble/appyx/core/composable/Children.kt
KovalevAndrey: added

documentation/ui/transitions.md
```diff
@@ -12,9 +12,133 @@ Below you can find the different options how to visualise `NavModel` state chang
...
 
-Using the provided [Child-related composables](children-view.md) you'll see no transitions as a default – UI changes resulting from the NavModel's state update will be rendered instantly. 
+Using the provided [Child-related composables](children-view.md) you'll see no transitions as a default – UI changes resulting from the NavModel's state update will be rendered instantly.
 
 
+## Shared element transitions
+
+To support shared element transition between two Child Nodes you need:
+
+1. Use sharedElementModifier with the same key on the composable you want to connect.
```
manuelvicnt:
```suggestion
1. Use the `sharedElement` Modifier with the same key on the composable you want to connect.
```

documentation/ui/transitions.md
```diff
@@ -12,9 +12,133 @@ Below you can find the different options how to visualise `NavModel` state chang
...
+   override fun View(modifier: Modifier) {
+      Box(
+         modifier = Modifier
+            // make sure you specify the size before using sharedElement modifier
+            .fillMaxSize()
+            .sharedElement(key = "sharedContainer")
+      ) { /** ... */ }
+   }
+}
+class NodeOne(
```
manuelvicnt:
```suggestion

class NodeTwo(
```

documentation/ui/transitions.md
```diff
@@ -12,9 +12,133 @@ Below you can find the different options how to visualise `NavModel` state chang
...
+      }
+   
+   @Composable
+   override fun View(modifier: Modifier) {
+      Children(
+         // or any other NavModel
+         navModel = backStack,
+         // or no transitionHandler at all. Using a slider will make the shared element slide away
+         // with the rest of of the content.
+         transitionHandler = rememberBackStackFader()
```
manuelvicnt:
```suggestion
         transitionHandler = rememberBackStackFader(),
         withSharedElementTransition = true
```

libraries/core/src/main/kotlin/com/bumble/appyx/core/composable/Children.kt
```diff
@@ -50,21 +59,49 @@ inline fun <reified NavTarget : Any, State> ParentNode<NavTarget>.Children(
...
+        }
+    } else {
+        Box(modifier = modifier
+            .onSizeChanged {
+                transitionBounds = it
+            }
+        ) {
+            CompositionLocalProvider(
+                /** If sharedElement is not supported for this Node - provide null otherwise children
+                 * can consume ascendant's  LocalSharedElementScope */
```
manuelvicnt:
```suggestion
                 * can consume ascendant's LocalSharedElementScope */
```

libraries/core/src/main/kotlin/com/bumble/appyx/core/navigation/transition/MovableContent.kt
```diff
@@ -0,0 +1,36 @@
...
+ * When this transition starts this function will return null for Node (A) and movable content for
+ * Node (B)so that this content will be moved from Node (A) to Node (B).
+ *
+ * If you have a custom NavModel keep in mind that you can only move content from a visible Node
+ * that becomes invisible to a Node that is becoming visible.
+ */
+@Composable
+fun localMovableContentWithTargetVisibility(
+    key: Any,
+    defaultValue: @Composable () -> Unit
```
manuelvicnt: Nit: Instead of `defaultValue`, this should be `content` right?

libraries/core/src/main/kotlin/com/bumble/appyx/core/navigation/transition/MovableContent.kt
```diff
@@ -0,0 +1,36 @@
...
+ * and we want to move content from Node (A) to Node (B). Node (A) is transitioning from Active to
+ * Stashed (invisible) state and Node (B) is transitioning from Created to Active (visible) state.
+ * When this transition starts this function will return null for Node (A) and movable content for
+ * Node (B)so that this content will be moved from Node (A) to Node (B).
+ *
+ * If you have a custom NavModel keep in mind that you can only move content from a visible Node
+ * that becomes invisible to a Node that is becoming visible.
+ */
+@Composable
+fun localMovableContentWithTargetVisibility(
```
manuelvicnt, line 26: We might want to create multiple `localMovableContentWithTargetVisibility` functions with different `content` lambdas as there are many `movableContentOf` functions you can call? https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/runtime/runtime/src/commonMain/kotlin/androidx/compose/runtime/MovableContent.kt

libraries/core/src/main/kotlin/com/bumble/appyx/core/node/ParentNode.kt
```diff
@@ -74,6 +76,15 @@ abstract class ParentNode<NavTarget : Any>(
         manageTransitions()
     }
 
+    @Composable
+    final override fun DerivedSetup(innerContent: @Composable () -> Unit) {
+        CompositionLocalProvider(
+            LocalParentNodeMovableContent provides mutableMapOf()
```
manuelvicnt: Should we create the `mutableMapOf` for every `ParentNode`? What if the `LocalParentNodeMovableContent` is set alongside `LocalSharedElementScope`? Is it possible? What are the implications?

This would effectively create a new map for every single `ParentNode` we have in the hierarchy, even if the Map isn't used

samples/sandbox/src/main/kotlin/com/bumble/appyx/sandbox/client/container/ContainerNode.kt
```diff
@@ -96,6 +101,7 @@ class ContainerNode internal constructor(
         when (navTarget) {
             is Picker -> node(buildContext) { modifier -> ExamplesList(modifier) }
             is NavModelExamples -> NavModelExamplesNode(buildContext)
+            is SharedElementFaderExample -> SharedElementWithMovableContentExampleNode(buildContext)
```
manuelvicnt: Should we add the two examples? Shared element transitions with and without movable content

samples/sandbox/src/main/kotlin/com/bumble/appyx/sandbox/client/sharedelement/FullScreenNode.kt
```diff
@@ -0,0 +1,60 @@
...
+        Box(
+            modifier = modifier
+                .fillMaxSize()
+                .clickable {
+                    onClick(profileId)
+                }
+        ) {
+            val profile = Profile.allProfiles[profileId]
+
+
```
manuelvicnt, line 84: Nit: extra line

manuelvicnt (commented):
Looks good to me! Added some comments but overall looks great :D thank you!

samples/sandbox/src/main/kotlin/com/bumble/appyx/sandbox/client/container/ContainerNode.kt
KovalevAndrey: I can add another one yeah

libraries/core/src/main/kotlin/com/bumble/appyx/core/navigation/transition/MovableContent.kt
KovalevAndrey: Supported other functions too

libraries/core/src/main/kotlin/com/bumble/appyx/core/node/ParentNode.kt
KovalevAndrey: Moved to Children  composable and add it when movableContent is enabled

documentation/ui/transitions.md
```diff
@@ -12,9 +12,135 @@ Below you can find the different options how to visualise `NavModel` state chang
...
+for this key if it doesn't exist. 
+
+In the example below when a NodeOne is being replaced with NodeTwo in a BackStack or Spotlight NavModel
+`CustomMovableContent("movableContentKey")` will be moved from NodeOne to NodeTwo without losing its
+state. 
+
+
+```kotlin
+@Composable
+fun CustomMovableContent(key: Any, modifier: Modifier = Modifier) {
```
manuelvicnt, line 117: Add the `ParentNode` to this example so that people know where to set `withMovableContent = true`?

manuelvicnt (approved):

PR #716, opened 2024-08-29
Update libraries

LachlanMcKee:
## Description
Updated Kotlin, Compose Compiler (implicitly via Kotlin) and several Androidx libs

## Check list


libraries/core/build.gradle.kts
```diff
@@ -52,6 +50,7 @@ dependencies {
     api(libs.androidx.appcompat)
     api(libs.kotlin.coroutines.android)
 
+    debugImplementation(libs.androidx.tracing) // Espresso fails without this
```
LachlanMcKee, line 53: It seems that one of the libraries uses tracing 1.0 instead of 1.1 which causes Espresso to fail with `No static method forceEnableAppTracing()V in class Landroidx/tracing/Trace`

libraries/core/src/main/kotlin/com/bumble/appyx/core/children/ChildNodeCreationManager.kt
```diff
@@ -49,7 +49,8 @@ internal class ChildNodeCreationManager<NavTarget : Any>(
                 when (keepMode) {
                     ChildEntry.KeepMode.KEEP -> {
                         navModelKeepKeys =
-                            (state.onScreen + state.offScreen).mapNotNullToSet { element -> element.key }
+                            state.onScreen.mapNotNullToSet { element -> element.key } +
```
LachlanMcKee, line 52: The Kotlin compiler reeeeeeally doesn't like `(state.onScreen + state.offScreen)`

I assume it has something to do with the generics, but even after attempting to change a lot of code to use `: Any` it still didn't work 🤔

libraries/core/src/main/kotlin/com/bumble/appyx/core/composable/Child.kt
```diff
@@ -38,8 +38,8 @@ fun <NavTarget : Any, State> ParentNode<NavTarget>.Child(
     transitionParams: TransitionParams,
     transitionHandler: TransitionHandler<NavTarget, State>,
     decorator: @Composable ChildTransitionScope<State>.(
-        child: ChildRenderer,
-        transitionDescriptor: TransitionDescriptor<NavTarget, State>
+        ChildRenderer,
+        TransitionDescriptor<NavTarget, State>
```
LachlanMcKee, line 42: This is no longer allowed in Kotlin 2.0
https://developer.android.com/jetpack/androidx/releases/compose-compiler#1.5.0

`Named arguments for @Composable lambda calls have been deprecated. This feature is relying on internal compiler APIs and will not be supported by K2.`

libraries/core/src/main/kotlin/com/bumble/appyx/core/children/ChildNodeCreationManager.kt
KovalevAndrey: those are 2 lists of the same type, looks like a compiler bug

LachlanMcKee: at least there was a workaround 🫠

samples/app/build.gradle.kts
```diff
@@ -79,9 +76,4 @@ dependencies {
     implementation(libs.compose.ui.tooling)
     implementation(libs.compose.ui.ui)
     implementation(libs.google.material)
-
-    androidTestImplementation(composeBom)
-    androidTestImplementation(libs.androidx.test.espresso.core)
-    androidTestImplementation(libs.androidx.test.junit)
-    androidTestImplementation(libs.compose.ui.test.junit4)
```
LachlanMcKee, line 86: There are no tests in this module (maybe there should be?)

libraries/core/src/main/kotlin/com/bumble/appyx/core/children/ChildNodeCreationManager.kt
KovalevAndrey: Does set + set create a new set or a collection?

LachlanMcKee: a new set.

the definition of `navModelKeepKeys` is `val navModelKeepKeys: Set<NavKey<NavTarget>>` so it wouldn't compile unless it returned a set.

KovalevAndrey (approved):

PR #718, opened 2024-11-18
Introduced global plugins, added GlobalNodeLifecycleAware

LachlanMcKee:
## Description
Added 'global plugins' to enable developers to introduce diagnostic tools without needing to find the RIB within the view hierarchy

## Check list

### DIFF
--- a/libraries/core/src/main/kotlin/com/bumble/appyx/core/plugin/Plugins.kt
+++ b/libraries/core/src/main/kotlin/com/bumble/appyx/core/plugin/Plugins.kt
@@ -23,6 +23,11 @@ interface NodeLifecycleAware : Plugin {
     fun onCreate(lifecycle: Lifecycle) {}
 }
 
+interface GlobalNodeLifecycleAware : Plugin {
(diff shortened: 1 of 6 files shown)


libraries/core/src/main/kotlin/com/bumble/appyx/core/plugin/Plugins.kt
```diff
@@ -23,6 +23,11 @@ interface NodeLifecycleAware : Plugin {
     fun onCreate(lifecycle: Lifecycle) {}
 }
 
+interface GlobalNodeLifecycleAware : Plugin {
```
CherryPerry, line 26: Can be implemented as combination of `NodeAware` and `NodeLifecycleAware`?

CherryPerry (approved):

libraries/core/src/main/kotlin/com/bumble/appyx/core/plugin/Plugins.kt
LachlanMcKee: As discussed in in person this won't work for the usecase.

